PR #10, opened 2024-07-04
timestamp_floor with offset updates

goodroot:
Per @nwoolmer:

There's a variant of this function with an extra parameter (https://github.com/questdb/questdb/pull/4692), so I've added a quick note about it.

src: https://github.com/questdb/questdb.io/pull/2355


goodroot:
@nwoolmer Merged, double-check? 🥗

PR #11, opened 2024-07-04
Backwards ASOF JOIN updates

goodroot:
Per @nwoolmer:

A short note about an optimisation added in https://github.com/questdb/questdb/pull/4612

https://github.com/questdb/questdb.io/pull/2354

### DIFF
--- a/reference/sql/join.md
+++ b/reference/sql/join.md
@@ -484,6 +484,44 @@ order, the join result is non-deterministic.
 
 :::
 
+### Backwards ASOF JOIN
+
+For the simple non-keyed case (i.e no ON clause), ASOF JOINs can be run in reverse, allowing for more performant access to recent data.
+
+DDL:
+
+```questdb-sql
+CREATE TABLE t1 (ts TIMESTAMP, i INT, s SYMBOL) timestamp(ts) partition by day bypass wal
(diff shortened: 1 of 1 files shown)


reference/sql/join.md
```diff
@@ -484,6 +484,44 @@ order, the join result is non-deterministic.
...
+| --------------------------- | --- | --- | --------------------------- | --- | --- |
+| 2023-10-06T01:00:00.000000Z | 7   | h   | 2022-10-10T01:01:00.000000Z | 3   | d   |
+| 2023-10-05T09:00:00.000000Z | 6   | g   | 2022-10-10T01:01:00.000000Z | 3   | d   |
+| 2022-10-05T08:19:00.000000Z | 5   | f   | 2022-10-05T08:17:00.000000Z | 1   | b   |
+| 2022-10-05T08:18:00.000000Z | 4   | e   | 2022-10-05T08:17:00.000000Z | 1   | b   |
+
+The above syntax is an alternative to:
+
+```questdb-sql
+SELECT * FROM t2
```
jerrinot, line 518: is this query working at all? I tried
```sql
(SELECT * FROM trades ORDER BY timestamp DESC)
ASOF JOIN
(SELECT * FROM trips ORDER BY pickup_datetime DESC);
```
on the demo and I'm getting this error: `left side of time series join doesn't have ASC timestamp order`
which makes sense to me. ASOF join assume both tables have monotonically increasing timestamps.

PR #27, opened 2024-07-24
Changes for SAMPLE BY FROM-TO

nwoolmer:
WIP Docs for https://github.com/questdb/questdb/pull/4733


reference/sql/sample-by.md
```diff
@@ -56,6 +58,71 @@ trades per hour:
...
+
+:::note
+
+This syntax extension was added in QuestDB 8.1.0, and therefore is unavailable for versions prior to this.
+
+Please see the new blog for more information.
+
+:::
+
+When using `SAMPLE BY` with `FILL`, missing rows within the result set can be filled with pre-determined values.
```
goodroot, line 74:
```suggestion
When using `SAMPLE BY` with `FILL`, you can fill missing rows within the result set with pre-determined values.
```

reference/sql/sample-by.md
```diff
@@ -56,6 +58,71 @@ trades per hour:
...
+
+However, this will only fill rows between existing data in the data set, and cannot be used to fill
```
goodroot, line 76:
```suggestion
However, this method will only fill rows between existing data in the data set and cannot fill rows outside of this range.
```

reference/sql/sample-by.md
```diff
@@ -56,6 +58,71 @@ trades per hour:
...
 ```
 
+
+## FROM-TO
+
+todo: blog link
+
+:::note
+
+This syntax extension was added in QuestDB 8.1.0, and therefore is unavailable for versions prior to this.
```
goodroot, line 68:
```suggestion
Versions prior to QuestDB 8.1.0 do not have access to this extension.
```

reference/sql/sample-by.md
```diff
@@ -56,6 +58,71 @@ trades per hour:
...
+When using `SAMPLE BY` with `FILL`, missing rows within the result set can be filled with pre-determined values.
+
+However, this will only fill rows between existing data in the data set, and cannot be used to fill
+rows outside of this range.
+
+To fill outside the bounds of the existing data, you can specify a fill range using a `FROM-TO` clause.
+
+#### Syntax
+
+The resulting shape of the query is specified using `FROM` and `TO`:
```
goodroot, line 83:
```suggestion
Specify the shape of the query using `FROM` and `TO`:
```

reference/sql/sample-by.md
```diff
@@ -56,6 +58,71 @@ trades per hour:
...
+
+```questdb-sql title='Pre-filling trip data' demo
+SELECT pickup_datetime as t, count
+FROM trips
+SAMPLE BY 1d FROM '2008-12-28' TO '2009-01-05' FILL(NULL)
+```
+
+Since there are no rows prior to 2009, these rows are filled automatically.
```
goodroot, line 91:
```suggestion
Since no rows existed before 2009, QuestDB automatically fills in these rows.
```

reference/sql/sample-by.md
```diff
@@ -56,6 +58,71 @@ trades per hour:
...
+
+This is distinct from the `WHERE` clause with a simple rule of thumb -
+`WHERE` controls what data flows in, `FROM-TO` controls what data flows out.
+
+Both `FROM` and `TO` can be used in isolation to solely pre-fill or post-fill data. If `FROM` is not provided,
+then the lower bound is the start of the dataset, aligned to calendar, and vice versa when `TO` is omitted.
```
goodroot, line 97:
```suggestion
Use both `FROM` and `TO` in isolation to pre-fill or post-fill data. If `FROM` is not provided, then the lower bound is the start of the dataset, aligned to calendar. The opposite is true omitting `TO`.
```

reference/sql/sample-by.md
```diff
@@ -56,6 +58,71 @@ trades per hour:
...
+
+#### `WHERE` clause optimisation
+
+If the user does not provide a `WHERE` clause, or the `WHERE` clause does not consider the designated timestamp,
+QuestDB will add one for you, matching the `FROM-TO` interval.
+
+This means that the query will run optimally, and avoid touching data not relevant to the result.
+
+Therefore, the prior query will be compiled into something similar to this:
```
goodroot, line 106:
```suggestion
Therefore, we compile the prior query into something similar to this:
```

reference/sql/sample-by.md
```diff
@@ -56,6 +58,71 @@ trades per hour:
...
+SAMPLE BY 1d FROM '2008-12-28' TO '2009-01-05' FILL(NULL)
+```
+
+#### Limitations
+
+Here are the current limits to this feature.
+
+- This syntax is not compatible with `FILL(PREV)` or `FILL(LINEAR)`.
+- This syntax is for `ALIGN TO CALENDAR` only (default alignment).
+- Any specified `OFFSET` will not be considered.
```
goodroot, line 122:
```suggestion
- Does not consider any specified `OFFSET`.
```

goodroot (commented):
Solid! We'll have to think through Railroad... 🚂 ...

Comments include examples of passive voice with alternatives.

It's tricky and seems pedantic. But it leads to better doc writing.

As always, take what you like.

reference/sql/sample-by.md
```diff
@@ -18,6 +18,8 @@ use of the [FILL](#fill-options) keyword to specify a fill behavior.
 
 ### SAMPLE BY keywords
 
+todo: railroad update
```
goodroot, line 21: We might be blocked here. A fix will take some thinking.

goodroot: @nwoolmer Agreed.

goodroot (approved):

PR #39, opened 2024-08-12
Update Kafka connect ports from 9009 -> 9000

nwoolmer:
The article uses port 9009 but with http addresses. Later, it uses http addresses with port 9000. I assume this should be consistently 9000.


goodroot (approved):
Indeed!

PR #42, opened 2024-08-14
chore(docs): add resume WAL troubleshooting

puzpuzpuz:


reference/sql/alter-table-resume-wal.md
```diff
@@ -51,16 +51,58 @@ The table `weather_wal` is suspended. The last successful commit in the table is
...
+
+## Diagnosting corrupted WAL transactions
+
+Sometimes a table may get suspended due to full disk or [kernel limits](/docs/deployment/capacity-planning/#os-configuration). In this case, a whole WAL segment may get corrupted. This means that there will be multiple transactions that rely on the corrupted segment and finding the transaction number to resume from may be complicated. When you run RESUME WAL on such suspended table, you may see an error like this one:
+
+```
+2024-07-10T01:01:01.131720Z C i.q.c.w.ApplyWal2TableJob job failed, table suspended [table=weather_wal~3, error=could not open read-only [file=/home/my_user/.questdb/db/weather_wal~3/wal45/101/_event], errno=2]
+```
+
+In such a case, you may try skipping all transactions that that rely on the corrupted WAL segment. To do that, first you need to find the last applied transaction number for the `weather_wal` table:
```
javier, line 76: extra 'that' on that sentence

reference/sql/alter-table-resume-wal.md
```diff
@@ -51,16 +51,58 @@ The table `weather_wal` is suspended. The last successful commit in the table is
 The following query restarts transactions from the failed transaction, `4`:
 
 ```questdb-sql
-ALTER TABLE  weather_wal RESUME WAL;
+ALTER TABLE weather_wal RESUME WAL;
```
javier, line 54: Could we please use the `trades` table in this whole page? That way we will be consistent with all the new examples. Thanks!

javier: We will need to edit the log lines in the docs, just to be extra sure we don't forget that

reference/sql/alter-table-resume-wal.md
```diff
@@ -51,16 +51,58 @@ The table `weather_wal` is suspended. The last successful commit in the table is
...
+  AND segmentId = 101;
+```
+
+Here, `1223` stands for the last applied transaction number, `45` stands for the WAL ID that may be seen in the error log above (`weather_wal~3/wal45`), and `101` stands for the WAL segment ID from the log (`weather_wal~3/wal45/101`).
+
+| max  |
+| ---- |
+| 1242 |
+
+Since the maximum number of the problematic transactions is `1242`, you can now run resume operation for the `1243` transaction:
```
javier, line 104: Could it be the case more than one consecutive transaction is wrong and you need to repeat the process until getting to the right point?

javier: Could we maybe add that the data at that particular transaction will be lost, but if you are using DEDUP and you have access to the original events (for example they are on Kafka, or other replayable source) it is safe to reingest?

javier (commented):
Please see my comments. Thanks

reference/sql/alter-table-resume-wal.md
puzpuzpuz: Fixed in bc9fbf09622226fff9e45637e8afd1a391b7596e

reference/sql/alter-table-resume-wal.md
puzpuzpuz: Done in bc9fbf09622226fff9e45637e8afd1a391b7596e

reference/sql/alter-table-resume-wal.md
puzpuzpuz: Yes, the process may have to be repeated. Added a trailing paragraph on that in bc9fbf09622226fff9e45637e8afd1a391b7596e

puzpuzpuz: Added a hint box with some words on dedup in e72fc14edf39da54f8796a3657e4eccea948fcfd

ideoma (approved):

puzpuzpuz:
@nwoolmer addressed all comments in c489b81fef5242d129ba965756200ae114239936. PTAL

PR #48, opened 2024-08-27
Add `hydrate_table_metadata` function

nwoolmer:
Related to https://github.com/questdb/questdb/pull/4848

Documents a new function which can be used to refresh the static metadata cache.

Shouldn't be closed until after the main issue is merged.


reference/function/meta.md
```diff
@@ -437,3 +437,45 @@ SELECT pg_catalog.version();
...
+## hydrate_table_metadata('table1', 'table2' ...)
+
+`hydrate_table_metadata' re-reads table metadata from disk to update the static metadata cache.
+
+:::warning
+
+This function should only be used when directed by QuestDB support. Misuse could cause corruption of the metadata 
+cache, requiring the database to be restarted.
+
+:::
```
goodroot, line 451: @nwoolmer This is spooky. 👻 

If the impact is dire, and it's by instruction, I'd lean towards a no-doc.

Is the potential scope of impact "only" that a DB restart is required?

goodroot (approved):

PR #62, opened 2024-10-08
interval type

nwoolmer:


reference/function/date-time.md
```diff
@@ -440,6 +440,83 @@ SELECT hour(ts), count() FROM transactions;
...
+two timestamps.
+
+**Arguments:**
+
+- `start_timestamp` is a timestamp.
+- `end_timestamp` is a timestamp.
+
+**Return value:**
+
+Return value type is `interval`
```
puzpuzpuz, line 455:
```suggestion
Return value type is `interval`.
```

reference/function/date-time.md
```diff
@@ -440,6 +440,83 @@ SELECT hour(ts), count() FROM transactions;
...
 
+## interval
+
+`interval(start_timestamp, end_timestamp)` - creates a time interval from
+two timestamps.
+
+**Arguments:**
+
+- `start_timestamp` is a timestamp.
+- `end_timestamp` is a timestamp.
```
puzpuzpuz, line 451:
```suggestion
- `end_timestamp` is a timestamp not earlier than the `start_timestamp`.
```

reference/function/date-time.md
```diff
@@ -736,6 +813,59 @@ SELECT second(ts), count() FROM transactions;
...
+
+```questdb-sql title="Using today"
+SELECT true as in_today FROM long_sequence(1)
+WHERE now() IN today();
+```
+
+
+## today, tomorrow, yesterday with timezone
+
+`today(tz)` - returns an interval representing the current day with timezone adjustment.
```
puzpuzpuz, line 841: We should probably use `timezone`, not `tz` here and below to keep things consistent with `to_timezone()` docs.

reference/function/date-time.md
```diff
@@ -736,6 +813,59 @@ SELECT second(ts), count() FROM transactions;
 | 58     | 9876  |
 | 59     | 2567  |
 
+
+## today, tomorrow, yesterday
+
+`today()` - returns an interval representing the current day.
```
puzpuzpuz, line 819: We should mention that the interval is in the UTC/GMT+0 timezone.

reference/function/date-time.md
```diff
@@ -736,6 +813,59 @@ SELECT second(ts), count() FROM transactions;
...
+`tz` is a `string` matching a timezone.
+
+**Return value:**
+
+Return value is of type `interval`.
+
+**Examples:**
+
+```questdb-sql title="Using today"
+
```
puzpuzpuz, line 856: Redundant newline here.

reference/function/date-time.md
```diff
@@ -736,6 +813,59 @@ SELECT second(ts), count() FROM transactions;
...
+No arguments taken.
+
+**Return value:**
+
+Return value is of type `interval`.
+
+**Examples:**
+
+```questdb-sql title="Using today"
+SELECT true as in_today FROM long_sequence(1)
```
puzpuzpuz, line 834: Maybe we should use `trades` or `trips` table instead of `long_sequence(1)`? The current example is a bit too artificial.

reference/function/date-time.md
```diff
@@ -736,6 +813,59 @@ SELECT second(ts), count() FROM transactions;
...
+
+**Return value:**
+
+Return value is of type `interval`.
+
+**Examples:**
+
+```questdb-sql title="Using today"
+
+SELECT today() as today, today('CEST') as adjusted
```
puzpuzpuz, line 857:
```suggestion
SELECT today() as today, today('CEST') as adjusted;
```

reference/operators/date-time.md
```diff
@@ -129,57 +186,35 @@ results on Jan 1-2 in 2018 and in 2019:
...
-The `BETWEEN` operator can also accept non-constant bounds. For instance, the
-following query returns all records older than one year from the current date:
+#### Examples
 
-```questdb-sql title="One year before current date"
-SELECT * FROM scores
-WHERE ts BETWEEN to_str(now(), 'yyyy-MM-dd')
-AND dateadd('y', -1, to_str(now(), 'yyyy-MM-dd'));
+```questdb-sql title="Check if timestamp is in interval success"
+SELECT true as is_in_interval FROM long_sequence(1)
```
puzpuzpuz, line 203: Again, would be nice to use a demo or a sample table here.

puzpuzpuz (approved):
LGTM. All comments are nits.

reference/function/aggregation.md
```diff
@@ -829,6 +833,68 @@ FROM (SELECT x FROM long_sequence(5));
 | :--------- |
 | 1,2,3,4,5  |
 
+## string_distinct_agg
+
+`string_distinct_agg(value, delimiter)` - Concatenates distinct non-null string
```
puzpuzpuz, line 838: nit: in docs for other functions we start with lower case
```suggestion
`string_distinct_agg(value, delimiter)` - concatenates distinct non-null string
```

reference/function/aggregation.md
```diff
@@ -829,6 +833,68 @@ FROM (SELECT x FROM long_sequence(5));
...
+values into a single string, using the specified delimiter to separate the
+values.
+
+- `string_distinct_agg` ignores null values and only concatenates non-null
+  distinct values.
+
+- The order of concatenated values is not guaranteed.
```
puzpuzpuz, line 845: The order is guaranteed - it's the scan order. So, if you're running `select string_distinct_agg(str_col, ',') from my_table;` and you have `a`, `b`, and `c` values in the `str_col` column, you'll get `a,b,c`.

reference/function/aggregation.md
```diff
@@ -829,6 +833,68 @@ FROM (SELECT x FROM long_sequence(5));
...
+
+### Arguments
+
+- `value`: A varchar or string column containing the values to be aggregated.
+- `delimiter`: A char value used to separate the distinct values in the
+  concatenated string.
+
+### Return value
+
+Return value type is `varchar`.
```
puzpuzpuz, line 855:
```suggestion
Return value type is `string`.
```

reference/function/aggregation.md
```diff
@@ -829,6 +833,68 @@ FROM (SELECT x FROM long_sequence(5));
...
+```
+
+This query will return a single string containing all the distinct skyCover
+values separated by commas. The skyCover column contains values such as OVC
+(Overcast), BKN (Broken clouds), SCT (Scattered clouds), and CLR (Clear skies).
+Even though the skyCover column may have many rows with repeated values,
+string_distinct_agg aggregates only the unique non-null values. The result is a
+comma-separated list of all distinct sky cover conditions observed.
+
+Result:
```
puzpuzpuz, line 873: The below result doesn't look like a comma-separated list of all distinct sky cover conditions, so it's a bit confusing. Also, there is no `Description` alias in the query.

goodroot: Was trying to be too helpful!

PR #65, opened 2024-10-16
Add regr_slope and table_storage

goodroot:


reference/function/finance.md
```diff
@@ -148,10 +152,160 @@ SELECT mid(1.5760, 1.5763)
...
+
+```questdb-sql
+SELECT regr_slope(y, x) AS slope FROM measurements;
+```
+
+Result:
+
+| slope |
+| ----- |
+| 8.0   |
```
mtopolnik, line 221: This should be 0.8

reference/function/finance.md
```diff
@@ -148,10 +152,160 @@ SELECT mid(1.5760, 1.5763)
...
+formula:
+
+$$
+b_1 = \frac{N \sum (xy) - \sum x \sum y}{N \sum (x^2) - (\sum x)^2}
+$$
+
+Where:
+
+- Let $N$ be the number of valid data points.
+- Let $\sum (xy)$ be the sum of the products of $x$ and $y$.
```
mtopolnik, line 181: These aren't "let" definitions, but explanations. I'd rephrase to "sum(xy) is the sum of ..."

reference/function/finance.md
```diff
@@ -148,10 +152,160 @@ SELECT mid(1.5760, 1.5763)
...
+```questdb
+SELECT regr_slope(y, x) AS slope FROM ( SELECT 1 AS x, 2 AS y UNION ALL SELECT
+2, NULL UNION ALL SELECT NULL, 4 UNION ALL SELECT 4, 5 );
+```
+
+Result:
+
+| slope |
+| ----- |
+| 8.0   |
```
mtopolnik, line 273: The slope between (1,2) and (4,5) is 1, so maybe this was also supposed to come out as 0.8?

goodroot: Great catch TY!

reference/function/finance.md
goodroot: I forgot math!

PR #82, opened 2024-11-25
Documentation updates for 8.2

goodroot:
* Window function improvements
* `regr_intercept()`
* `line.tcp.acl.enabled`

### DIFF
--- a/documentation/reference/function/window.md
+++ b/documentation/reference/function/window.md
@@ -245,7 +245,17 @@ Frame boundaries determine which rows are included in the window calculation:
 - `<value> PRECEDING`: Starts or ends at a specified number of rows or interval before the current row
 - `CURRENT ROW`: Starts or ends at the current row
 
-> **Default Frame**: If unspecified, defaults to `RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`
+When the frame clause is not specified, the default frame is
+`RANGE UNBOUNDED PRECEDING`, which includes all rows from the start of the
+partition to the current row.
+
+- If `ORDER BY` is not present, the frame includes the entire partition, as all
+  rows are considered equal.
+
+- If `ORDER BY` is present, the frame includes all rows from the start of the
+  partition to the current row. Note that `UNBOUNDED FOLLOWING` is only allowed
+  when the frame start is `UNBOUNDED PRECEDING`, which means the frame includes
+  the entire partition.
 
 ### Restrictions
 
@@ -276,15 +286,36 @@ Modifies the window frame by excluding certain rows:
 
 ### avg()
 
-Calculates the average of a numeric expression over the window frame.
+In the context of window functions, `avg(value)` calculates the average of
+`value` over the set of rows defined by the window frame.
+
+**Arguments:**
+
+- `value`: The column of numeric values to calculate the average of.
+
+**Return value:**
+
+- The average of `value` for the rows in the window frame.
+
+**Description**
+
+When used as a window function, `avg()` operates on a "window" of rows defined
+by the `OVER` clause. The rows in this window are determined by the
+`PARTITION BY`, `ORDER BY`, and frame specification components of the `OVER`
+clause.
+
+The `avg()` function respects the frame clause, meaning it only includes rows
+within the specified frame in the calculation. The result is a separate average
+for each row, based on the corresponding window of rows.
+
+Note that the order of rows in the result set is not guaranteed to be the same
+with each execution of the query. To ensure a consistent order, use an
+`ORDER BY` clause outside of the `OVER` clause.
 
 **Syntax:**
 ```questdb-sql title="avg() syntax" 
 avg(value) OVER (window_definition)
 ```
-
-**Arguments:**
-- `value`: Numeric column or expression
 
 **Example:**
 ```questdb-sql title="avg() example" demo
@@ -300,17 +331,38 @@ SELECT
 FROM trades;
 ```
 
-### sum()
+### Cumulative sum()
 
-Calculates the sum of a numeric expression over the window frame.
+In the context of window functions, `sum(value)` calculates the sum of `value`
+in the set of rows defined by the window frame. Also known as "cumulative sum".
+
+**Arguments:**
+
+- `value`: Any numeric value.
+
+**Return value:**
+
+- The sum of `value` for the rows in the window frame.
+
+**Description**
+
+When used as a window function, `sum()` operates on a "window" of rows defined
+by the `OVER` clause. The rows in this window are determined by the
+`PARTITION BY`, `ORDER BY`, and frame specification components of the `OVER`
+clause.
+
+The `sum()` function respects the frame clause, meaning it only includes rows
+within the specified frame in the calculation. The result is a separate value
+for each row, based on the corresponding window of rows.
+
+Note that the order of rows in the result set is not guaranteed to be the same
+with each execution of the query. To ensure a consistent order, use an
+`ORDER BY` clause outside of the `OVER` clause.
 
 **Syntax:**
 ```questdb-sql title="sum() syntax" 
 sum(value) OVER (window_definition)
 ```
-
-**Arguments:**
-- `value`: Numeric column or expression
 
 **Example:**
 ```questdb-sql title="sum() example" demo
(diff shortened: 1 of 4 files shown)


documentation/reference/function/window.md
```diff
@@ -74,173 +79,189 @@ some order, like the three cars who finished before and after a certain car.
...
 
-##### Common Syntax
+This query demonstrates different time intervals in action, calculating:
+- 5-minute moving average of best bid price
+- Update frequency in 100ms windows
+- 2-second rolling volume
 
-Both `RANGE` and `ROWS` modes share similar syntax for defining the frame:
+:::note
+RANGE frames require ORDER BY on a numeric or timestamp column.
```
javier, line 255: This would be nice, but unfortunately it is not true. As of now RANGE will complain if using any order by different to the designated timestamp, unless I am doing this wrong

javier: Also, related to this, at the moment RANGE behaves wrongly and we have an open issue for this, but when using RANGE, all the rows with the same value will have the same output for the function, as explained in this issue https://github.com/questdb/questdb/issues/5177. We should probably document it, even if it is broken and needs fixing

documentation/reference/function/window.md
```diff
@@ -525,220 +498,25 @@ Note that the order of rows in the result set is not guaranteed to be the same
...
-| ETH-USD | 2615.35  | 0.02245868 | 2022-03-08T18:03:58.000000Z | 2615.54     |
-| ETH-USD | 2615.36  | 0.03244613 | 2022-03-08T18:03:58.000000Z | 2615.54     |
-| BTC-USD | 39265.27 | 0.00006847 | 2022-03-08T18:03:58.000000Z | 39269.98    |
-| BTC-USD | 39262.42 | 0.00046562 | 2022-03-08T18:03:58.000000Z | 39269.98    |
-
-#### First value of price in the latest second, descending designated timestamp order
-
-```questdb-sql
+**Example:**
+```questdb-sql title="first_value() example" demo
```
javier, line 507: We have an opportunity to add here a second example (or maybe under the common examples section) for a LAG function replacement, as we don't have one today

```
SELECT
    symbol,
    price,
    timestamp,
    first_value(price) OVER (
        PARTITION BY symbol        
        ORDER BY timestamp
        ROWS 1 PRECEDING EXCLUDE CURRENT ROW
    ) AS price_from_1_row_before,
    first_value(price) OVER (
        PARTITION BY symbol        
        ORDER BY timestamp
        ROWS 2 PRECEDING EXCLUDE CURRENT ROW
    ) AS price_from_2_rows_before
FROM trades;
```

documentation/reference/function/window.md
```diff
@@ -856,300 +595,200 @@ Note that the order of rows in the result set is not guaranteed to be the same
...
+  - Based on logical intervals
+  - Require ORDER BY on timestamp or numeric column
+  - Support time-based intervals (e.g., '1h', '5m')
+
+### Exclusion behavior
+- Using `EXCLUDE CURRENT ROW` with frame end at `CURRENT ROW`:
+  - Automatically adjusts end boundary to `1 PRECEDING`
+  - Ensures consistent results across queries
+
+### Window-only function restrictions
```
javier, line 757: as commented earlier, I would remove this classification and these restrictions

documentation/reference/function/window.md
```diff
@@ -856,300 +595,200 @@ Note that the order of rows in the result set is not guaranteed to be the same
...
+### Performance considerations
+- ROWS frames typically perform better than RANGE frames for large datasets
+- Partitioning can improve performance by processing smaller chunks of data
+- Consider index usage when ordering by timestamp columns
+
+### Common pitfalls
+
+#### Using window functions in WHERE clauses:
+
+```questdb-sql title="Not allowed!"
```
javier, line 774: We should explain how to do this, which is by enclosing the query in a subquery using WITH, then adding a query with the WHERE, as in
```
with prices_and_avg AS (
SELECT
    symbol,
    price, avg(price) OVER (ORDER BY timestamp) as moving_avg_price,
    timestamp
FROM trades
WHERE timestamp in yesterday()
)
select * from prices_and_avg
WHERE
   moving_avg_price  > 100;
```

documentation/reference/function/window.md
```diff
@@ -856,300 +595,200 @@ Note that the order of rows in the result set is not guaranteed to be the same
...
-| ETH-USD | 2615.35  | 0.02245868 | 2022-03-08T18:03:58.000000Z | 5230.89   |
-| ETH-USD | 2615.36  | 0.03244613 | 2022-03-08T18:03:58.000000Z | 7846.25   |
-| BTC-USD | 39265.27 | 0.00006847 | 2022-03-08T18:03:58.000000Z | 235594.46 |
-| BTC-USD | 39262.42 | 0.00046562 | 2022-03-08T18:03:58.000000Z | 274856.88 |
-
-#### Moving price sum over whole result set
-
-```questdb-sql
+#### Missing ORDER BY in OVER clause
+
```
javier, line 786: We should explain what is the potential issue here. Actually this query is perfectly fine and it works as expected, and in the car example we had above this is what we were suggesting to do. When no ORDER BY is specified, the avg will be calculated for the whole partition, in this case since we don't have a partition by and we are using a global window, all the rows would show the same average, which is the avg for the whole dataset (I would add a where timestamp in yesterday() so it can be executed on demo).

Then we can show the same query adding the ORDER BY, and we will see now we have a global moving average.

```
SELECT
    symbol,
    price,
    sum(price) OVER (ORDER BY TIMESTAMP) AS cumulative_sum
FROM trades
WHERE timestamp in yesterday();
```
Or we can also show a query in which we have PARTITION BY and no ORDER BY, so all the rows for the same partition (symbol) will have the same AVG
```
SELECT
    symbol,
    price,
    sum(price) OVER (PARTITION BY symbol ) AS cumulative_sum
FROM trades
WHERE timestamp in yesterday();
```
And we can finally show the query with both partition by and order by, so every row shows the moving avg for each symbol

```
SELECT
    symbol,
    price,
    sum(price) OVER (PARTITION BY symbol ORDER BY TIMESTAMP) AS cumulative_sum
FROM trades
WHERE timestamp in yesterday();
```

javier (changes requested):
There are a few things I would change. But looking much better than before and the diagrams and query examples are a super nice touch

documentation/reference/function/window.md
```diff
@@ -74,173 +75,189 @@ some order, like the three cars who finished before and after a certain car.
...
-        RANGE BETWEEN '1' HOUR PRECEDING AND CURRENT ROW)
-as moving_avg
-FROM trades
+```txt
+ROWS BETWEEN 2 PRECEDING AND CURRENT ROW
 ```
 
-This calculates the average price for each symbol, for the current row and all
-rows with a timestamp within the preceding hour.
+This includes the current row and four preceding rows.
```
javier, line 170: TWO preceding rows, as we changed the query :)

documentation/reference/function/window.md
```diff
@@ -254,31 +271,82 @@ partition to the current row.
...
-  with the same `ORDER BY` value in `RANGE` mode. This is equivalent to setting
-  the frame end to `1 PRECEDING`.
-- `EXCLUDE NO OTHERS`: Includes all rows in the frame. This is the default if no
-  exclusion option is specified.
+2. Frame end can only be:
+   - `CURRENT ROW`
+   - `<value> PRECEDING` (unless start is `UNBOUNDED PRECEDING`)
 
-### Time Units
+3. RANGE frames must have ORDER BY on a timestamp or numeric column
```
javier, line 285: that would be ideal, as it is the standard, but on questdb as of today range functions need to be ordered by designated timestamp

documentation/reference/function/window.md
```diff
@@ -856,300 +591,289 @@ Note that the order of rows in the result set is not guaranteed to be the same
...
-| ETH-USD | 20000   | 0.1        | 1       |
-| BTC-USD | 1479.45 | 0.02       | 2       |
-| ETH-USD | 20000   | 0.000249   | 2       |
+Key Points:
 
-## Cumulative sum
+- `LAG(price, 1)` fetches the price from the immediately preceding row for the same symbol.
+- `PARTITION BY symbol` ensures the function operates independently for each stock.
+- `ORDER BY timestamp` processes rows in chronological order.
+- `price - LAG(price, 1)` calculates the price change compared to the previous row.
```
javier, line 735: We are not doing that calculation on this query. We are getting the value from the row before and from 2 rows before, but we are not doing the substraction here

documentation/reference/function/window.md
```diff
@@ -856,300 +591,289 @@ Note that the order of rows in the result set is not guaranteed to be the same
...
-`ORDER BY` clause outside of the `OVER` clause.
+- ROWS frames:
+  - Based on physical row counts
+  - More efficient for large datasets
+  - Can be used with any ORDER BY column
 
-**Examples:**
+- RANGE frames:
+  - Based on logical intervals
+  - Require ORDER BY on timestamp or numeric column
```
javier, line 781: as of today, we only support order by designated timestamp

javier (dismissed):
looking good. Just a few minor things

PR #84, opened 2024-11-26
Fix typo in quick-start

sleto-it:
`find installation` instead of `installation find`


goodroot (approved):
😆 I stared at this for a good minute before seeing what was missing. Ahh, typos...

Thanks!

PR #86, opened 2024-11-28
docs for `DECLARE` syntax

nwoolmer:
Docs for https://github.com/questdb/questdb/pull/5207

Also contains missing updates for `ATOMIC` (`CREATE`, `INSERT`) and `FROM`-`TO` for `SAMPLE BY`.


javier (commented):
Please, add trailing semicolons to all sample queries, so when it opens on a web console (or if I copy and paste) I can execute directly even if I have other queries already there

goodroot (approved):

PR #87, opened 2024-11-29
clarify DEDUP sorting behaviour

nwoolmer:
@goodroot  Perhaps this should be a section with a heading and not a note? It ended up larger than expected.


goodroot (approved):

goodroot:
Note works. :)

PR #136, opened 2025-02-14
Fix SAMPLE BY doc typos

puzpuzpuz:


puzpuzpuz:
> Happy with the changes, fyi you can use `count` without `()` and it still executes in QuestDB SQL. Not sure how 'intended' it is!

The typo was in "rows outside of this range. rows outside of this range.", not in SQL. SQL changes are merely to make the samples consistent.

puzpuzpuz:
@nwoolmer thanks for the review!

PR #150, opened 2025-03-14
Document interval expressions supported by timestamp_floor() SQL function

puzpuzpuz:


puzpuzpuz:
@nwoolmer thanks for the review!

PR #172, opened 2025-04-28
Document SQL hints

jerrinot:
Adding docs for SQL hints introduced in 8.3.1
Preview link for lazy reviewers: https://preview-172--questdb-documentation.netlify.app/concept/sql-optimizer-hints/
It's cross-linked from the ASOF Join doc page: https://preview-172--questdb-documentation.netlify.app/reference/sql/asof-join/#sql-performance-hints-for-asof-join


documentation/concept/sql-hints.md
```diff
@@ -0,0 +1,133 @@
+---
+title: SQL Hints
```
amunra, line 2: On the team call I was confused on what hints were about.
You then mentioned that Oracle has these so I googled it and saw "optimizer hints" and it immediately clicked what these were about.

Maybe a title rename?

amunra: https://docs.oracle.com/cd/E11882_01/server.112/e41573/hintsref.htm#PFGRF005

documentation/concept/sql-hints.md
```diff
@@ -0,0 +1,133 @@
...
+description:
+  SQL Hints allow expert users to guide the query optimizer in QuestDB when default optimization strategies are not optimal.
+  This document describes available hints and when to use them.
+---
+
+QuestDB's query optimizer automatically selects execution plans for SQL queries based on heuristics. However, in some
+scenarios, the optimizer may not have all the information needed to choose the most efficient plan. SQL Hints provide
```
amunra, line 9: Devil's advocate.. it's got the hardware info, utilization stats and actual data at hand.. what more could it need?

Minor rewording suggested so it doesn't look too bad :-)

documentation/concept/sql-hints.md
```diff
@@ -0,0 +1,133 @@
...
+SELECT /*+ HINT_NAME(parameter1 parameter2) */ columns FROM table;
+```
+
+Hints are entirely optional and designed to be a safe optimization mechanism:
+
+- The database will use default optimization strategies when no hints are provided
+- Syntax errors inside a hint block won't fail the entire SQL query
+- The database safely ignores unknown hints
+- Only block comment hints (`/*+ HINT */`) are supported, not line comment hints (`--+ HINT`)
+
```
amunra, line 27: Question.. how would one debug these? If I use a hint that does not exist, will it error?

jerrinot: Hints are designed to avoid failing a query. It must be safe to execute a query with a hint on a questdb version which does not support the hint. It's a comment after all, a comment must not fail a query. This is perhaps the only place where I do prefer NOT to fail-fast. 


See this: https://preview-172--questdb-documentation.netlify.app/concept/sql-hints/#execution-plan-observation
this is specific to the `USE_ASOF_BINARY_SEARCH`, but that's only mechanism we have now. @nwoolmer had an idea that EXPLAIN could indicate that an operator was picked because of a hint, but that's currently not implemented.

documentation/concept/sql-hints.md
jerrinot: that's a very good one, thanks!

documentation/concept/sql-hints.md
jerrinot: do you have any suggestion?

documentation/sidebars.js
```diff
@@ -489,6 +489,7 @@ module.exports = {
         },
         "concept/root-directory-structure",
         "concept/sql-extensions",
+        "concept/sql-hints",
```
amunra, line 492: `sql-optimizer-hints` for better seo?

jerrinot: possibly, yes. let me change it.

documentation/concept/sql-hints.md
amunra: How about: 

based on heuristics. The default execution strategy should be the fastest for most datasets. You may use hints to select a specific execution strategy which may (or may not) outperform the default strategy. SQL hints influence the execution strategy of queries without changing their semantics.

documentation/concept/sql-hints.md
amunra: Once again I feel we should be able to send back warnings from the server for these sorts of things..

A syntax error in a hint would be quite annoying to debug.

jerrinot: indeed, that would be nice.

in fact, [pgwire](https://www.postgresql.org/docs/current/protocol-flow.html) has a facility for this: 'NOTICE' messages. but the rest of our infrastructure is not connected to this. 
> **NoticeResponse**
> A warning message has been issued. The frontend should display the message but continue listening for ReadyForQuery or ErrorResponse.

jerrinot: @amunra is this good to resolve? since you are asking for a new feature rather than a docs improvement.

amunra: Yeah, no worries.
Up to you if you want to provide more details on debugging, e.g. logs.

jerrinot:
@amunra thank you for reviewing it, it's better thanks to your feedback!

PR #173, opened 2025-04-29
Document TTL columns in tables() meta function

puzpuzpuz:


amunra (approved):
Approving,

That said (and not restricted to these) the tables on this page require quite a lot of horizontal scrolling to read.

Not sure what we can do about it.

PR #184, opened 2025-05-14
cross-links pgwire guides

jerrinot:
also:
1. removed dups/outdated info
2. typos (psychopg -> psycopg)


goodroot (approved):
Pretty sure Psychop is a pokemon 🤔

PR #269, opened 2025-10-30
Update supported datatypes for first and last agg functions

mcadariu:
Follow-up of https://github.com/questdb/questdb/pull/6291 and https://github.com/questdb/questdb/pull/6322.


nwoolmer (approved):
Thank you for the contribution(s)!

PR #277, opened 2025-11-12
Update partitions.md (Time Partitions > Properties > Hour)

SanderLedegen:
Hello team!
I was checking the documentation about [partitioning](https://questdb.com/docs/concept/partitions/#properties) and came across a mismatch between the docs and my QuestDB instance (9.1.1).

In short: the partition format for hourly partioned tables is not `YYYY-MM-DD-HH` as specified in the docs but turns out to be `YYYY-MM-DDTHH`. The UI running on port 9000 also confirms this:

afbeelding

Also, on the disk, in my quest db root folder where the database files reside, the same format is used (so with a `T`).


nwoolmer (approved):
Good catch, thanks for the contribution!

PR #296, opened 2025-12-17
docs: add WINDOW JOIN SQL reference documentation

bluestreak01:
## Summary

- Add comprehensive documentation for WINDOW JOIN, a new SQL join type for time-based aggregation
- Document syntax, RANGE clause, INCLUDE/EXCLUDE PREVAILING behavior
- Include 8 practical examples covering common use cases
- Add to navigation under Reference → SQL → SELECT → WINDOW JOIN

## Test plan

- [x] Verify page renders correctly with `yarn start`
- [x] Check navigation placement in sidebar
- [x] Validate all code examples are properly formatted

### DIFF
--- a/documentation/guides/architecture/time-series-optimizations.md
+++ b/documentation/guides/architecture/time-series-optimizations.md
@@ -72,7 +72,7 @@ sequential reads, materialized views, and in-memory processing.
 
 - Materialized views can be chained, with the output of one serving as the input to another, and support TTLs for lifecycle management.
 
-### Time-To_Life and Data lifecycle
+### Time-To-Live and Data lifecycle
 
 QuestDB supports [Time To Live (TTL)](/docs/concept/ttl/) configuration for both regular tables and
 materialized views. With TTL enabled, partitions older than the configured horizon will automatically


documentation/reference/sql/window-join.md
```diff
@@ -0,0 +1,269 @@
...
+   and be partitioned
+2. The slave (right) table must be a direct table reference, not a subquery
+3. Aggregate functions are required - you cannot select non-aggregated columns
+   from the slave table
+4. Symbol-based join conditions enable "Fast Join" optimization when matching on
+   symbol columns
+
+## Supported aggregate functions
+
+WINDOW JOIN supports the following aggregate functions on the slave table:
```
puzpuzpuz: We support all aggregate functions in WINDOW JOIN, not only these ones. Yet, the listed functions (plus `last` and `first_not_null`/`last_not_null`) use faster internal API (`computeBatch` a.k.a. "SIMD-optimized aggregation" in this text), so when only they're used, the query will run faster.

documentation/reference/sql/window-join.md
```diff
@@ -0,0 +1,269 @@
...
+    t.ts,
+    sum(p.price) AS future_sum
+FROM trades t
+WINDOW JOIN prices p
+    ON (t.sym = p.sym)
+    RANGE BETWEEN 1 minute FOLLOWING AND 2 minutes FOLLOWING
+    EXCLUDE PREVAILING;
+```
+
+### Cross-table aggregation (no symbol match)
```
puzpuzpuz: nit: shall we also add an example showing chained WINDOW JOINs?

bluestreak01: Added in 1631e16. Added a 'Chained WINDOW JOINs' example showing how to chain multiple WINDOW JOINs with different tables and time windows.

documentation/reference/sql/window-join.md
```diff
@@ -0,0 +1,275 @@
...
+2. The slave (right) table must be a direct table reference, not a subquery
+3. Aggregate functions are required - you cannot select non-aggregated columns
+   from the slave table
+4. Symbol-based join conditions enable "Fast Join" optimization when matching on
+   symbol columns
+
+## Aggregate functions
+
+WINDOW JOIN supports all aggregate functions on the slave table. However, the
+following functions use an optimized internal API and will run faster:
```
puzpuzpuz: nit: later on we refer to this as "SIMD-optimized aggregation", so
```suggestion
following functions use SIMD-optimized aggregation and will run faster:
```

documentation/reference/sql/window-join.md
```diff
@@ -0,0 +1,299 @@
...
+
+`UNBOUNDED PRECEDING` and `UNBOUNDED FOLLOWING` are not supported in WINDOW
+JOIN.
+
+:::
+
+### INCLUDE/EXCLUDE PREVAILING
+
+- `INCLUDE PREVAILING` (default): Includes slave rows within the time window
+  plus the most recent slave row before the window start, useful for "last known
```
puzpuzpuz: > within the time window plus the most recent slave row before the window start

This is misleading: the behavior is the same as for ASOF JOIN, so we add the most recent row that has the same or earlier timestamp to the start of the time window.

PR #426, opened 2026-05-01
docs(sql): SAMPLE BY FILL - cross-column references for prev, fast-path and limitations

jovfer:
## Summary

Refresh the SAMPLE BY FILL documentation to reflect what the fast-path implementation in [questdb/questdb#6946](https://github.com/questdb/questdb/pull/6946) now accepts.

- `query/sql/sample-by.md`
  - Add `PREV(col)` to the FILL options table and list its constraints (type match, no broadcast, no SYMBOL).
  - Replace the outdated "fill strategies cannot be combined" rule with the actual `FILL(NONE)` rule.
  - Rewrite FROM-TO Limitations: drop "not compatible with FILL(PREV)" and "does not consider OFFSET"; document that keyed FROM-TO works when bounds are constants and is rejected for bind variables, function calls, or arithmetic expressions.
- `cookbook/sql/time-series/fill-keyed-arbitrary-interval.md`
  - Scope the recipe to the bind-variable / function-bound case, since constant-bound keyed FROM-TO is now supported natively.
- `cookbook/sql/time-series/fill-from-one-column.md`
  - Replace the "imagine this SQL was valid" framing with the actual `FILL(PREV(col))` syntax. Keep the window-function rewrite as the flexible alternative (e.g. flagging filled rows).

## Test plan

- Read the rendered pages locally and confirm the new restrictions and examples line up with the behaviour exercised by the FILL test suite.
- Spot-check the linked anchors (`#from-to`, `#fill-options`) still resolve.


documentation/query/sql/sample-by.md
```diff
@@ -226,7 +230,7 @@ ALIGN TO CALENDAR TIME ZONE 'Europe/Berlin';
 Here, `FROM '2026-01-01T00:00:00'` means midnight January 1st in Berlin local time
 (= `2025-12-31T23:00:00Z` UTC).
 
-`FROM-TO` can only be used on non-keyed SAMPLE BY queries (queries with no grouping columns other than the timestamp).
+`FROM-TO` works with both non-keyed and keyed `SAMPLE BY` queries. Keyed queries emit one row per (bucket, key) combination across the full interval. Keyed `FROM-TO` requires `FROM`/`TO` to be constants; bind variables, function calls, or arithmetic expressions are rejected with _FROM-TO intervals are not supported for keyed SAMPLE BY queries_.
```
javier: I think there is some punctuation problem here. I am not sure which the restrictions are

documentation/query/sql/sample-by.md
```diff
@@ -247,10 +251,12 @@ SAMPLE BY 1d FROM '2009-01-01' TO '2009-01-10' FILL(NULL);
 
 ### Limitations
 
-- Not compatible with `FILL(PREV)` or `FILL(LINEAR)`.
+- Not compatible with `FILL(LINEAR)`.
 - Only works with `ALIGN TO CALENDAR` (default alignment).
-- Does not consider any specified `OFFSET`.
-- Only for non-keyed `SAMPLE BY` (designated timestamp and aggregate columns only). See [Fill keyed queries with arbitrary intervals](/docs/cookbook/sql/time-series/fill-keyed-arbitrary-interval/) for a workaround.
+- Keyed `FROM-TO` requires `FROM`/`TO` to be constants. Bind variables,
```
javier, line 256: This is exactly the same we said above. I would keept it here under limitations, and would remove from aboive

javier (changes requested):
Small comments, but otherwise ready to go

documentation/query/sql/sample-by.md
jovfer: it's poor formatting, anyway I'll delete the 3rd sentence here to avoid duplication with limitation sections as suggested below.

javier (approved):

PR #431, opened 2026-05-07
fix invalid range for timestamp_ns

RaphDal:
This PR fixes the TIMESTAMP_NS range, it was documented to be 2920 years while it actually is 292 years.


javier (approved):

javier:
thanks!

PR #449, opened 2026-05-15
Document regr_r2 coefficient of determination aggregate

jovfer:
Pairs with [questdb/questdb#7104](https://github.com/questdb/questdb/pull/7104) — the implementation PR for the new `regr_r2(y, x)` aggregate.

## Summary

- Add a `regr_r2` section to the finance functions reference, alphabetically between `regr_intercept` and `regr_slope`.
- Add a corresponding entry to the May 2026 reference changelog.

## What the section covers

- One-line definition (coefficient of determination, 0-1 scale).
- Behavior bullets: null when X is constant (covers the single-row case), 1.0 when Y is constant and X varies (per SQL:2003), supported types, argument order, and the divergence from `corr(y, x)` at the constant-Y edge.
- Math: the centered-moments form $r^2 = S_{xy}^2 / (S_{xx} \cdot S_{yy})$, with a note on the $S_{xx} = 0$ and $S_{yy} = 0$ branches.
- Three worked examples mirroring the surrounding sections:
  - A fleet-telemetry trend-detection query that uses `regr_slope` plus `regr_r2 > 0.7` as a significance filter.
  - The shared `measurements` table, with $r^2 = 0.81$.
  - A `GROUP BY` example over `sales_data` showing strong per-category linear fit.
- A null-handling example confirming that pairs with either argument null are skipped.

## Tradeoffs and notes

- The page reuses the existing `measurements` and `sales_data` example tables to keep the regression section coherent. The R-squared values for those tables are reported as computed, not hand-picked for visual appeal.
- The divergence from `corr()` at $S_{yy} = 0$ is called out explicitly. Users who reach for `regr_r2` expecting strict mathematical equivalence with `corr^2` will notice this one boundary; the alternative (returning NaN there) would diverge from the SQL standard.
- The fleet-telemetry example references `motor_temp` and `elapsed_seconds` columns to match the roadmap motivation (questdb/roadmap#120 gap 3). The query is illustrative; the page does not provide a runnable demo table for it.
- This PR documents only `regr_r2`. The other six standard regression aggregates listed in the same roadmap gap (`regr_count`, `regr_avgx`, `regr_avgy`, `regr_sxx`, `regr_sxy`, `regr_syy`) are not implemented yet and not documented here.

## Test plan

- Build the docs locally (`yarn start`) and visit `/docs/query/functions/finance/#regr_r2` to verify rendering.
- Confirm KaTeX renders the centered formula correctly.
- Verify the changelog entry links to the new anchor.
- Cross-check the worked-example values against the implementation in [questdb/questdb#7104](https://github.com/questdb/questdb/pull/7104).


documentation/query/functions/finance.md
```diff
@@ -340,6 +340,143 @@ Result:
...
+linear fit).
+
+- The function requires at least two valid (y, x) pairs to compute a value.
+  - If fewer than two pairs are available, the function returns null.
+- The function returns null when x is constant across all rows (zero variance
+  in the independent variable). This includes the single-row case.
+- The function returns 1 when y is constant and x varies (a horizontal line
+  fits constant y perfectly). This follows the SQL:2003 specification and
+  differs from `corr(y, x)`, which returns null in that case.
+- Supported data types for x and y include `double`, `float`, and `integer`
```
javier: Nit, but we are mentioning params all the time as y and x, so maybe we can resort here?

javier (changes requested):
Minor nit with the order and y and x, but we need to change the SQL examples

javier (approved):

PR #452, opened 2026-05-22
Document twap() designated-timestamp limitation

mtopolnik:
`twap()` now rejects any timestamp argument other than the table's
designated timestamp. This updates the `twap()` reference page so the
`timestamp` parameter description states that requirement, framed as a
current limitation.

Companion to the questdb engine change that adds the rejection.


documentation/query/functions/aggregation.md
```diff
@@ -1659,9 +1659,10 @@ Supports `SAMPLE BY` with `FILL` modes.
 #### Parameters
 
 - `price` is any numeric value.
-- `timestamp` is a `timestamp` value. This is typically the table's
-  [designated timestamp](/docs/concepts/designated-timestamp/) but can be any
-  timestamp column.
+- `timestamp` must be the table's
+  [designated timestamp](/docs/concepts/designated-timestamp/). As a current
+  limitation, `twap()` rejects a query that passes any other timestamp column
```
javier: I would refrain from saying as a current limitation. We can just say it rejects a query that passes any other timestamp column. We can add a link to https://questdb.com/docs/query/functions/timestamp/ if the user needs to see how to convert a non-designated timestamp column  into the designated timestamp for the query

javier (approved):

PR #462, opened 2026-06-05
Document per-query and per-principal memory limits

puzpuzpuz:
## Summary

Documents the new memory-limit feature across configuration, monitoring, and access control.

**OSS — per-workload memory limits**
- `configuration/cairo-engine.md`: new "Memory limits" section covering the reloadable workload limits (`cairo.query.memory.limit.bytes`, `cairo.mat.view.refresh.memory.limit.bytes`, `cairo.wal.apply.memory.limit.bytes`, plus a pointer to `cairo.live.view.refresh.memory.limit.bytes`, which the live-views config page already documents), per-workload breach behavior, and a note that memory-mapped memory does not count toward a limit. Cross-links added from the WAL, materialized-views, and capacity-planning pages.
- `query/functions/meta.md`: `query_activity` gains the `memory_used` and `memory_limit` columns.
- `configuration/materialized-views.md`: documents `cairo.mat.view.refresh.busy.retry.limit` and `cairo.mat.view.refresh.busy.retry.timeout`, which govern what happens after a refresh breaches its memory limit and were previously undocumented. The breach description in `cairo-engine.md` links to them.

**Enterprise — per-principal memory limits**
- `query/sql/acl/alter-user.md`, `alter-service-account.md`, and a new `alter-group.md` (+ sidebar): `SET MEMORY LIMIT { <size> | UNLIMITED }`.
- `security/rbac.md`: new "Memory limits" section — how limits resolve, what a per-principal limit does and does not cover, how to inspect limits, the `SET MEMORY LIMIT` permission, and `ALTER GROUP` in the command list.
- `query/sql/show.md`: the `memory_limit` column on `SHOW USERS`, `SHOW GROUPS` and `SHOW GROUPS <user>`, and `SHOW SERVICE ACCOUNTS` and `SHOW SERVICE ACCOUNTS <user | group>`.

**Limit resolution — override, not min.** Limits resolve by strict precedence: a principal's own limit → (users only) the most-restrictive limit among its groups → the global `cairo.query.memory.limit.bytes` workload limit. A `0`/`UNLIMITED` value at a level means "not set" and falls through to the next. A more specific limit, when set, fully **overrides** the broader one and **binds even when larger**, so a per-user or per-group override can raise a principal's ceiling above the workload limit, not only lower it. Service accounts never inherit group limits. (Earlier drafts of these docs described a most-restrictive "min" behavior; revised per questdb/questdb-enterprise@fb4e050 "Make per-entity memory limits override, not min".)

## Corrections made after checking against the implementation

Two statements in earlier revisions of this PR were wrong, and both were verified against the Enterprise code rather than re-reasoned:

- **`UPDATE` was described as never picking up a per-principal limit.** False for a non-WAL table. `UpdateOperation.apply()` passes the caller's own `SqlExecutionContext` to `UpdateOperatorImpl.executeUpdate`, which calls `queryRegistry.register(...)`, which acquires a QUERY-workload tracker from that context's security context — so a non-WAL `UPDATE` **is** capped by `SET MEMORY LIMIT`. An operator sizing one from the old text would have met an unexpected OOM abort. Only the WAL path (the default table type) runs under the WAL apply job's tracker. Replaced with a "What a per-principal limit covers" section that splits the two.
- **`SHOW GROUPS <user>` and `SHOW SERVICE ACCOUNTS <user | group>` were shown without a `memory_limit` column**, with prose saying so explicitly. They carry it — `AbstractShowLinkedEntitiesCursorFactory` adds it to both metadata shapes, pinned by `ShowAclTest`. Since these are among the results that break a positional client, the omission hid half the breaking change.

Example tables that disagreed with each other on the same entity's limit were also made consistent.

## Behaviour the docs previously did not mention at all

- **`COPY ... TO` exports are capped by the issuing principal's limit.** The export runs under that principal's own context on the query workload, so a limit must be sized for the largest single thing the principal runs, exports included — not only for its interactive queries.
- **`SET MEMORY LIMIT` is self-escalating.** It takes no entity name, so its holder can set *any* principal's limit including its own; combined with override-not-min, a non-admin holding it can raise its own ceiling above `cairo.query.memory.limit.bytes`. Documented as a warning, together with the `GRANT ALL` expansion — a principal granted `ALL` before the upgrade does not acquire the new permission.
- **The breaking change.** Five results gain a column on upgrade (`SHOW USERS`, `SHOW GROUPS` and `SHOW GROUPS <user>`, `SHOW SERVICE ACCOUNTS` and `SHOW SERVICE ACCOUNTS <user | group>`), plus `SELECT *` on `sys.acl_entities`. Positional clients must be updated; name-based clients are unaffected.
- **`sys.acl_entities`** stores the value, and is readable only by the built-in admin — a principal holding `DATABASE ADMIN` is still denied.
- **`memory_limit` means different things per statement**: the *effective* limit in `SHOW USERS`, the entity's *own* limit in `SHOW GROUPS` / `SHOW SERVICE ACCOUNTS`, since neither inherits one.
- **An external (SSO/OIDC) user's inherited group limit refreshes at next login**, not on its current session — the same refresh-on-login model that already governs group-granted permissions for external users.

## Re-validation against the merged code

Both source PRs merged after the last revision of this doc PR (OSS on 2026-06-20, Enterprise on 2026-09-08), and OSS `master` moved on afterwards. Every claim was re-checked against merged Enterprise `main` and OSS `master` (`questdb` submodule pin `fef82615`), and these were stale:

- **There are four workload limits, not three.** Live views (questdb/questdb#7461) added `LIVE_VIEW_REFRESH` as a workload with its own reloadable `cairo.live.view.refresh.memory.limit.bytes`. The live-views config page on `main` already documents the key, so the memory-limits section now points to it rather than duplicating it, and the intro, capacity-planning blurb, and the RBAC "not covered by a principal limit" list name live view refresh.
- **A mat view refresh breach no longer invalidates the view on first breach.** questdb/questdb#7275 reclassified a refresh OOM as a transient, retried failure: the refresh is deferred by `cairo.mat.view.refresh.busy.retry.timeout` and the view is invalidated only after `cairo.mat.view.refresh.busy.retry.limit` consecutive failures (`MatViewRefreshJob.isRetriableRefreshError`, pinned by `WorkloadMemoryTrackerTest`). A live view refresh breach, by contrast, invalidates immediately (`LiveViewRefreshJob`). The breach paragraph is now a per-workload list.
- **`ALTER USER ... SET MEMORY LIMIT` on an external (SSO/OIDC) user is rejected** (`AccessListModel.setMemoryLimit`: `Cannot set memory limit for external user`). The docs previously only said such users inherit from groups; they now say the direct form is refused, on both `rbac.md` and `alter-user.md`.
- **Coverage note trimmed.** The earlier "coverage is best-effort" paragraph, with its list of tracked allocation sites, is replaced by the one fact an operator needs: memory-mapped memory, such as the column files a query reads, does not count toward a limit.
- **WAL apply limit rarely fires.** WAL apply runs only simple `UPDATE`s, metadata ALTERs, and commits, so the entry now says the limit mainly isolates WAL apply from the query budget.
- **Upgrade note.** The `memory_limit` column is added by an automatic migration (`SysMig6`) on the first primary boot or on promote; an ACL statement refused with a message naming the `memory_limit` column migration means the node is not yet migrated. The window normally closes once WAL apply catches up, so retry first; a persistent refusal calls for a restart, which re-runs the migration.
- `query_activity`'s two memory columns are `null` for nested registrations, which share the outer query's budget.

Confirmed unchanged and still accurate: the `SET MEMORY LIMIT` grammar and permission name, override-not-min precedence and `0`/`UNLIMITED` semantics (`EntMemoryTrackerProvider.effectiveLimit`), the `memory_limit` column on all five `SHOW` results including the filtered forms and the `enabled`/`external_alias`/`grant_option` neighbours (`AbstractShowEntitiesCursorFactory`, `AbstractShowLinkedEntitiesCursorFactory`, `ShowAclTest`), the `COPY ... TO` and non-WAL `UPDATE` coverage, the breach message format (`Unsafe.checkPerQueryAllocLimit`), reloadability of every limit (`DynamicPropServerConfiguration`), and the built-in admin being unlimitable.

## Dependencies

Documents:
- questdb/questdb#7184 — per-query / per-workload memory limits (OSS), merged
- questdb/questdb-enterprise#1039 — per-principal memory limits (Enterprise), merged

Web-console syntax highlighting for the new clause: questdb/sql-parser#27

## Verification

`yarn build`, which this repo configures to **throw** on broken links, broken markdown links, and broken anchors.

## Review follow-up

A level-2 `/review-pr` pass (six agents, findings verified against OSS `fef82615` and Enterprise `main`) found no wrong technical claims. Its major findings are fixed in 4be3ffac: stale `SHOW SERVICE ACCOUNTS` tables on the grant and revoke assume-service-account pages, the `OUT OF MEMORY` error tag literal and the `wal_tables()` `errorTag`/`errorMessage` columns, the `queryId` mapping for materialized view refreshes, `CREATE GROUP ... WITH EXTERNAL ALIAS`, the two migration-window refusal messages, the RBAC page description, and the `query_activity` column order with the missing `is_wal` column.


documentation/concepts/live-views.md
```diff
@@ -236,6 +236,9 @@ base columns its query references:
 - Dropping, renaming, or changing the type of a referenced column invalidates the
   view.
 - Renaming or dropping the base table invalidates the view.
+- Exceeding
+  [`cairo.live.view.refresh.memory.limit.bytes`](/docs/configuration/live-views/#cairoliveviewrefreshmemorylimitbytes)
```
javier, line 240: Is it possible to add some way of figuring out the amount of memory that will be used?

Asking as getting a suspended view due to memory allocation can be a poor experience for the users. On a mat view you get invalidation only after a user action (or if server ran out of disk), but in this case it can be tricky for users. I know default is no limit, but maybe we could have some indication of how to calculate an appropriate size

javier: Actually, just noticed mat views also can get invalidated due to memory XD

documentation/concepts/materialized-views.md
```diff
@@ -454,6 +454,11 @@ modified in incompatible ways:
 - Renaming the base table
 - `TRUNCATE` or `UPDATE` operations
 
+A view is also invalidated when its refresh keeps failing with an out-of-memory
+error, including a breach of the
+[refresh memory limit](/docs/configuration/cairo-engine/#memory-limits), after
```
javier, line 459: Same as with live views. Do we have any way of giving an indication of the memory it would be needed?

documentation/concepts/live-views.md
javier: And just learnt query_activity gives you memory_used, so maybe we can have an example of how to calculate the memory used by a live view by running the query that the view would run with some limit of rows?

documentation/concepts/write-ahead-log.md
```diff
@@ -122,6 +122,8 @@ WAL behavior can be tuned via server configuration:
 
 - `cairo.wal.enabled.default` — WAL enabled by default (default: `true`)
 - Parallel threads for WAL application — see [WAL configuration](/docs/configuration/wal/)
+- `cairo.wal.apply.memory.limit.bytes` — cap on the native memory a WAL apply
+  batch may allocate; see [memory limits](/docs/configuration/cairo-engine/#memory-limits)
```
javier: for some reason I cannot see this on the preview. Not sure if some issue with markdown?

documentation/configuration/cairo-engine.md
```diff
@@ -87,6 +92,41 @@ A global timeout in seconds for long-running queries. When `query.timeout` is
 also set, it takes precedence and this key is ignored. Per-query overrides work
 the same as for `query.timeout`.
 
+### ram.usage.limit.bytes
```
javier: should these two properties now be part of the memory limits section?

They are already cross referenced from that section, so maybe they are fine here. Just to double check

documentation/query/sql/acl/alter-group.md
```diff
@@ -0,0 +1,85 @@
...
+  LDAP group alias. Applies to RBAC in QuestDB Enterprise."
+---
+
+import { EnterpriseNote } from "@site/src/components/EnterpriseNote"
+
+<EnterpriseNote>
+  RBAC provides fine-grained database permissions management.
+</EnterpriseNote>
+
+`ALTER GROUP` caps the query memory of a group's members with
```
javier: Do we need this intro about the two features here? For comparison, alter user (which has more options) has a minimal intro and then it details what each syntax block can do in the description

documentation/query/sql/acl/create-group.md
```diff
@@ -25,10 +25,24 @@ see the [RBAC operations](/docs/security/rbac) page.
 CREATE GROUP [IF NOT EXISTS] groupName;
 ```
 
+```questdb-sql title="Create a group mapped to an external group"
+CREATE GROUP groupName WITH EXTERNAL ALIAS externalAlias;
+```
+
 ## Description
 
 `CREATE GROUP` adds a new user group with no permissions.
```
javier, line 34: I think we should link here to the alter group memory limits section, since it cannot be set on creation directly

documentation/query/sql/acl/alter-user.md
```diff
@@ -2,8 +2,8 @@
 title: ALTER USER reference
 sidebar_label: ALTER USER
 description:
-  "ALTER USER SQL keywords reference documentation. Applies to RBAC in QuestDB
-  Enterprise."
+  "ALTER USER enables or disables a user, manages passwords and tokens, and sets
```
javier, line 5: The create user page should link in the description to this page, mentioning if you want to change memory limits, do it at the user level or group level via ALTER

documentation/query/sql/acl/alter-group.md
javier: Both the create group and create user pages should link in their descriptions to this page, mentioning if you want to change memory limits, do it at the user level or group level via ALTER

documentation/query/sql/acl/alter-service-account.md
```diff
@@ -2,8 +2,9 @@
 title: ALTER SERVICE ACCOUNT reference
 sidebar_label: ALTER SERVICE ACCOUNT
 description:
-  "ALTER SERVICE ACCOUNT SQL keywords reference documentation.  Applies to RBAC
-  in QuestDB Enterprise."
+  "ALTER SERVICE ACCOUNT enables or disables a service account, manages
```
javier, line 5: The create service account page should link in the description to this page, mentioning if you want to change memory limits, do it via ALTER

documentation/query/sql/alter-table-resume-wal.md
```diff
@@ -33,18 +33,24 @@ the Sequencer. Once the error is resolved, `ALTER TABLE RESUME WAL` restarts the
 suspended WAL transactions from the failed transaction. Alternatively, an
 optional `sequencerTxn` value can be provided to skip the failed transaction.
 
+`wal_tables()` also reports the `errorTag` and `errorMessage` of a suspended
+table; `OUT OF MEMORY` means a WAL apply batch breached its
```
javier: shouldn't this note be also on the alter mat view resume wal page above?

documentation/query/sql/update.md
```diff
@@ -25,6 +25,10 @@ SET columnName = expression [, columnName = expression ...]
   [attached by a symbolic link](/docs/query/sql/alter-table-attach-partition/#symbolic-links),
   the partition is read-only. `UPDATE` operation on a read-only partition will
   fail and generate an error.
+- On a WAL table, `UPDATE` is applied by the WAL apply job and counts against
```
javier, line 28: Is this true for all shapes of the update command, including when the update includes a select statement?

documentation/security/rbac.md
```diff
@@ -141,6 +140,17 @@ GRANT ILP TO ingest_app;                    -- InfluxDB Line Protocol access
 GRANT INSERT ON sensor_data TO ingest_app;  -- Can only insert into sensor_data
 ```
 
+### Cap a noisy tenant's query memory
+
+Bound the native memory each query from a group may allocate, so one tenant's
```
javier: This is for groups, but the title says "cap a noisy tenant's query memory", so I would expect to cap a single user. I would change the example to a user, and leave the link we already have to the memory limits section for other scenarios

documentation/security/rbac.md
```diff
@@ -578,6 +591,209 @@ information without these permissions.
...
+  `SHOW GROUPS`, and `SHOW SERVICE ACCOUNTS` require `LIST USERS`; the filtered
+  forms require `USER DETAILS` unless the caller names itself or one of its own
+  groups. A user without `LIST USERS` can still read its own effective cap from the
+  `memory_limit` column of `query_activity`, which always lists the caller's
+  own queries.
+- The stored value is persisted on the `sys.acl_entities` system table. That
+  table is protected: only the built-in admin can read it, and an ACL principal
+  holding `DATABASE ADMIN` is still denied.
+
+### Upgrading {#memory-limit-upgrade}
```
javier, line 754: We should probably mention in which version we are introducing these limits, so users don't need to figure it out

documentation/concepts/live-views.md
puzpuzpuz: There is no up-front estimate: what a refresh allocates depends on the data it reads, the plan the engine picks, and the parallelism, so the only way is to measure it.

I added a **Sizing a limit** subsection to the memory limits section of the Cairo engine page ([here](https://github.com/questdb/documentation/blob/c34ca900/documentation/configuration/cairo-engine.md#sizing-a-limit)) with the two-session `query_activity` recipe you suggested: run the query with the limit at `0` and sample `memory_used` from a second session. It is a live gauge with no peak value, so you poll and take the largest sample as the floor.

It then covers the two background workloads, which don't show up in `query_activity` but run SQL you can reproduce:
- a mat view refresh runs the view's `SELECT` restricted to the refreshed time range, so you run that `SELECT` with a `WHERE` on the designated timestamp spanning one refresh worth of data (whole table if you need full refreshes to fit);
- a live view refresh runs the window functions over each batch of new rows, so the same proxy measures the transient part; the retained state (`in_mem_bytes` from `live_views()` plus window state) is added on top, and the allocation-floor note on the config key still applies.

It also says that after a breach the `used` and `size` values in the message are the record of the footprint. The invalidation bullets on both concept pages now link to the subsection. c34ca900

documentation/concepts/materialized-views.md
puzpuzpuz: Same answer as for live views: it has to be measured. The new **Sizing a limit** subsection ([here](https://github.com/questdb/documentation/blob/c34ca900/documentation/configuration/cairo-engine.md#sizing-a-limit)) describes running the view's `SELECT` over one refresh worth of data and sampling `memory_used` in `query_activity`, and this invalidation paragraph now links to it. c34ca900

documentation/configuration/cairo-engine.md
puzpuzpuz: Moved both keys into the Memory limits section, after the three workload limits, so everything memory-related is in one place; the section intro was already introducing them. Anchors come from the heading text, so existing links to `#ramusagelimitbytes` and `#ramusagelimitpercent` are unchanged. c34ca900

documentation/query/sql/acl/alter-group.md
puzpuzpuz: Trimmed it to `ALTER GROUP modifies group settings.`, matching ALTER USER. The details were already in Description. c34ca900

documentation/query/sql/acl/create-group.md
puzpuzpuz: Added a paragraph to the Description: a new group has no limit (`SHOW GROUPS` reports `null`), with a link to `ALTER GROUP ... SET MEMORY LIMIT`. c34ca900

documentation/query/sql/acl/alter-user.md
puzpuzpuz: Added to the CREATE USER Description: it cannot set a limit, with links to `ALTER USER ... SET MEMORY LIMIT` for the user level and `ALTER GROUP ... SET MEMORY LIMIT` for the group level, plus the RBAC memory limits section for how the two interact. c34ca900

documentation/query/sql/acl/alter-group.md
puzpuzpuz: Done for both: CREATE GROUP links to `ALTER GROUP ... SET MEMORY LIMIT`, and CREATE USER links to both `ALTER USER` and `ALTER GROUP`. c34ca900

documentation/query/sql/acl/alter-service-account.md
puzpuzpuz: Added to the CREATE SERVICE ACCOUNT Description, linking to `ALTER SERVICE ACCOUNT ... SET MEMORY LIMIT`, with a note that service accounts do not inherit group limits. c34ca900

documentation/query/sql/alter-table-resume-wal.md
puzpuzpuz: Yes. Added it under "Check WAL status" on the ALTER MATERIALIZED VIEW RESUME WAL page.

While there I made the wording on both pages precise: `OUT OF MEMORY` is set for any out-of-memory failure of a WAL apply batch, whether against `cairo.wal.apply.memory.limit.bytes` or the process-wide `ram.usage.limit.*` (`ApplyWal2TableJob` derives the tag from `CairoException.isOutOfMemory()`). The mat view page additionally separates this from a refresh breaching `cairo.mat.view.refresh.memory.limit.bytes`, which invalidates the view instead of suspending it, so `RESUME WAL` does not help there. c34ca900

documentation/query/sql/update.md
puzpuzpuz: Yes. For a WAL table the compiler never runs the `UPDATE` on the caller's thread: every accepted form, including one with a subquery in `SET` or `WHERE`, is written to the WAL as SQL and executed by `ApplyWal2TableJob` under the WAL apply tracker (`SqlCompilerImpl.compileUpdate` only builds an `UpdateOperation` for non-WAL tables or when called from WAL application).

The one shape that never reaches the WAL is `UPDATE ... FROM` (a join): it is rejected on WAL tables with `UPDATE statements with join are not supported yet for WAL tables`. The note now states both. c34ca900

documentation/security/rbac.md
puzpuzpuz: Changed the example to `ALTER USER tenant_a SET MEMORY LIMIT 1G;` and the sentence to a user's queries. The link to the Memory limits section stays for the group and service account cases. c34ca900

documentation/security/rbac.md
puzpuzpuz: Added. The RBAC section now says per-principal limits are available since QuestDB Enterprise 4.0.2 and that the workload limits they override are available since QuestDB 10.0.0; the Cairo engine memory limits section states 10.0.0 as well.

How I got the numbers: OSS questdb/questdb#7184 merged on 2026-06-20, after the 9.4.3 tag (Jun 15) and before the 10.0.0 release (Aug 6); Enterprise 4.0.0's `questdb` pin (10.0.1) contains it. Enterprise questdb/questdb-enterprise#1039 merged on 2026-09-08, after the 4.0.1 tag (Sep 1), and `main` is at 4.0.2-SNAPSHOT. If the next Enterprise release ends up numbered differently, that one string needs updating. c34ca900

javier (approved):

PR #468, opened 2026-06-15
docs: add EXPIRE ROWS materialized-view documentation

nwoolmer:
Deep-dive concept page (concepts/deep-dive/expire-rows.md) for passthrough materialized views with EXPIRE ROWS: all modes, worked examples, read-filter/cleanup mechanics, NULL/ties/monotonicity semantics, and limitations. Adds an ALTER MATERIALIZED VIEW SET EXPIRE reference page, an EXPIRE ROWS section in CREATE MATERIALIZED VIEW, and sidebar entries.


mtopolnik:
Verdict: approve with nits

All four blocking items are fixed and verified. yarn build passes clean.

Resolved:

- trades schema drift — fixed, and more thoroughly than I flagged: create-mat-view.md:84-92 was also realigned to (symbol, side, price, amount, timestamp), which I hadn't caught. I re-derived all four worked-example result tables in expire-rows.md against the new INSERT data — the added side values are correct in every row (trades_sized, trades_latest, trades_peak, trades_top2), including the two where the side of the surviving row isn't the obvious one (trades_peak BTC→sell, trades_top2 ETH 50.0→sell).
- data-retention hub — fixed. Five bullets, count matches the prose, and the caveat ("Some policies hide expired rows without reclaiming their disk space") is the right thing to say on a page whose stated purpose is saving disk.
- show.md CLEANUP EVERY — fixed, and you propagated it to expire-rows.md:602-604 ("It omits CLEANUP EVERY when the cadence is the default 1h"), which matches expireToSink:99-103 exactly.
- trades_mirror undefined — fixed with a CREATE plus a prose line naming the columns.

One thing I checked because the new trades_mirror example omits it: passthrough views inherit the base table's partitioning when PARTITION BY is absent (CreateMatViewOperationImpl:989-996), and every MatViewExpireRowsTest case creates them that way. So all the bare CREATE MATERIALIZED VIEW ... AS (SELECT * FROM trades) EXPIRE ROWS ... examples are valid as written.

Still open, all non-blocking and your call — the Minors from last round are unchanged (sidebars.js:346 alphabetical order, "the top-N form keeps exactly one row per group" at expire-rows.md:46, the truncated KEEP 2 HIGHEST price ... at :624, the missing backlink in alter-mat-view-set-ttl.md, "summarising" at materialized-views.md:18, the meta.md whitespace churn, 'BTC' vs demo BTC-USDT, description length).

One new, introduced by the rename: materialized-views.md:214 is now the only surviving bare ts in the changed pages — LATEST ON ts PARTITION BY sensor_id, on a sensor_readings table whose schema the page never gives. It read as fine when everything else used ts; now it reads as a stray. Either name the column timestamp or give sensor_readings a one-line schema.

The two claims I couldn't verify last time are unchanged and still worth a confirm from you: the Parquet-compaction-reverts-to-native note (expire-rows.md:678-681) and the Enterprise primary-only cleanup claim (:435-437).

documentation/sidebars.js
```diff
@@ -633,6 +634,7 @@ module.exports = {
             "concepts/deduplication",
             "concepts/delivery-semantics",
             "concepts/ttl",
+            "concepts/expire-rows",
```
javier, line 637: the sidebar says EXPIRE ROWS, which is not nice when seen together with the rest of concepts, which are not in uppercase

documentation/changelog.mdx
```diff
@@ -8,6 +8,23 @@ description: Recent updates and improvements to the QuestDB documentation.
 
 This page tracks significant updates to the QuestDB documentation.
 
+## Unreleased
```
javier: This cannot go like this in the changelog, as we will later for sure not updated. When we publish, it should just say the month in which these docs were changed. The changelog is for the docs themselves, not for questdb, so nothing goes into an unreleased category

documentation/changelog.mdx
```diff
@@ -8,6 +8,23 @@ description: Recent updates and improvements to the QuestDB documentation.
...
+
+### New
+
+- [Expiring rows (`EXPIRE ROWS`)](/docs/concepts/expire-rows/) and [ALTER MATERIALIZED VIEW SET EXPIRE](/docs/query/sql/alter-mat-view-set-expire/) - New row-level retention for materialized views, with creation syntax in [CREATE MATERIALIZED VIEW](/docs/query/sql/create-mat-view/), immediate read filtering, and background reclamation for eligible policies
+
+### Compatibility
```
javier: Did the skill add this category??? we use standard categories for the changelog, and this is not one of them

documentation/concepts/expire-rows.md
```diff
@@ -0,0 +1,729 @@
...
+Once the policy takes effect, ordinary queries hide expired rows in every mode
+**without waiting for physical cleanup**. This does not mean synchronous ALTER
+completion or immediate invalidation of dependent views. A background job
+reclaims their on-disk storage afterwards under a monotonic
+`WHEN` predicate; the relative modes (`KEEP LATEST`, `KEEP HIGHEST/LOWEST`,
+`KEEP N`) and window predicates hide rows without freeing disk. See
+[The modes](#the-modes) and
+[Monotonicity and cleanup safety](#monotonicity-and-cleanup-safety).
+
+## The modes
```
javier: should this title just read Modes?

documentation/concepts/expire-rows.md
```diff
@@ -0,0 +1,729 @@
...
+
+Every mode keeps a defined set of rows and expires the rest. A row is expired
+only when the rule selects it for removal.
+
+| Mode                  | What it keeps                                       | Syntax                                                              | Frees disk            |
+| --------------------- | --------------------------------------------------- | ------------------------------------------------------------------- | --------------------- |
+| Per-row predicate     | Rows for which the predicate is **not** `TRUE`      | `EXPIRE ROWS WHEN predicate`                                        | Yes, when monotonic   |
+| Keep latest           | The latest row per key (current state per key)      | `EXPIRE ROWS KEEP LATEST [ON timestamp] PARTITION BY cols`           | No (read filter only) |
+| Keep highest / lowest | Rows tied at the group max / min of a column        | `EXPIRE ROWS KEEP HIGHEST\|LOWEST col [PARTITION BY cols]`          | No (read filter only) |
```
javier: I know this is not just docs, but I find it weird that LATEST is ON and HIGHEST/LOWEST do not use ON. A bit of a UX friction

documentation/concepts/expire-rows.md
```diff
@@ -0,0 +1,729 @@
...
+
+`DROP EXPIRE` does not automatically restore an invalidated dependent. After
+resolving the source conflict, materialized views require
+[FULL refresh](/docs/query/sql/refresh-mat-view/#full); live views require saving
+the definition and [recreation](/docs/concepts/live-views/#base-table-lifecycle).
+FULL refresh retains its general truncate-first failure behavior; see the linked
+reference for visibility during rebuilding and failure recovery. Rebuilding can
+only recover source rows that still exist.
+
+## Worked examples
```
javier: Not sure this is a great title section. Why "Worked"?

documentation/concepts/expire-rows.md
```diff
@@ -0,0 +1,729 @@
...
+
+A predicate on the designated timestamp gives a **rolling retention window**,
+which is the main use for `WHEN`. It is re-evaluated on every read, so the
+visible set rolls forward with the clock, and no `WHERE` clause can express it
+because the defining query rejects `now()`:
+
+```questdb-sql title="Keep the last 1 day"
+CREATE MATERIALIZED VIEW trades_recent AS (
+  SELECT * FROM trades
+) EXPIRE ROWS WHEN timestamp < dateadd('d', -1, now());
```
javier, line 263: hard pass! all new examples need to use TICK syntax, unless not possible

documentation/concepts/expire-rows.md
```diff
@@ -0,0 +1,729 @@
...
+```
+
+(With the 2024 sample timestamps above, every row is already older than a day
+and would be hidden; use recent data to see rows retained.)
+
+`WHEN timestamp < dateadd('d', -1, now())` and
```
javier, line 269: TICK

documentation/concepts/expire-rows.md
```diff
@@ -0,0 +1,729 @@
...
+| BTC    | sell | 105.0 | 2.0    | 2024-01-01T11:00:00.000000Z |
+| BTC    | buy  | 102.0 | 1.5    | 2024-01-02T09:00:00.000000Z |
+| ETH    | buy  | 55.0  | 1.0    | 2024-01-02T08:00:00.000000Z |
+| ETH    | sell | 50.0  | 3.0    | 2024-01-01T10:30:00.000000Z |
+
+BTC keeps its two highest (`105`, `102`) and drops `100`; ETH has only two rows,
+so both survive. Ties are broken by the designated timestamp, so the N-th
+boundary is deterministic.
+
+### Window predicate: the escape hatch
```
javier: Not a big fan of using "escape hatch". Maybe we can use something around "generic" or "custom" which would be easier to reason about?

documentation/concepts/expire-rows.md
```diff
@@ -0,0 +1,729 @@
...
+## `WHERE` filter or `EXPIRE ROWS`?
+
+A passthrough view can exclude rows in two places: a `WHERE` clause in its
+defining query, or an `EXPIRE ROWS WHEN` predicate. The dividing line is the
+clock.
+
+**Use `EXPIRE ROWS WHEN` for rules that move with wall-clock time.** A rolling
+window cannot be written as a `WHERE` clause at all: a view's defining query
+rejects non-deterministic functions, so
+`WHERE timestamp > dateadd('d', -7, now())` is not accepted.
```
javier: ok, maybe good use of dateadd here, but can we also be explicit about what happens with TICK syntax, as it is our preferred syntax for time ranges?

documentation/query/sql/alter-mat-view-set-expire.md
```diff
@@ -0,0 +1,144 @@
...
+`trades` table with `symbol`, `side`, `price`, `amount`, and designated
+`timestamp` columns:
+
+```questdb-sql title="Create the passthrough view"
+CREATE MATERIALIZED VIEW trades_mirror AS (SELECT * FROM trades);
+```
+
+```questdb-sql title="Rolling 7-day window, with a tighter cleanup cadence"
+ALTER MATERIALIZED VIEW trades_mirror
+  SET EXPIRE ROWS WHEN timestamp < dateadd('d', -7, now()) CLEANUP EVERY 30m;
```
javier, line 80: TICK syntax

documentation/query/sql/alter-mat-view-set-expire.md
```diff
@@ -0,0 +1,144 @@
...
+`SET EXPIRE` is allowed when materialized or live views already depend on this
+view. Those dependents detect an applied policy when they next refresh and
+become invalid; ALTER completion and dependent invalidation are separate events.
+An idle dependent may remain active, and an already-running refresh may finish
+against its earlier snapshot. Existing dependent rows are not retroactively
+filtered. Removing the policy does not automatically reverse invalidation. See
+[Dependent materialized and live views](/docs/concepts/expire-rows/#dependent-materialized-and-live-views)
+for consequences and recovery.
+
+## Examples
```
javier, line 68: There is no example of WHEN with an OVER clause

documentation/query/sql/create-mat-view.md
```diff
@@ -278,6 +316,62 @@ Time units: `HOURS`, `DAYS`, `WEEKS`, `MONTHS`, `YEARS`
...
+remove a policy with
+[`ALTER MATERIALIZED VIEW SET EXPIRE`](/docs/query/sql/alter-mat-view-set-expire/).
+
+A view can carry both `TTL` and `EXPIRE ROWS`. `TTL` comes first in the
+statement and first in effect: it removes rows from the view, and the
+`EXPIRE ROWS` policy then applies to the rows that stay. See
+[Combining with TTL](/docs/concepts/expire-rows/#combining-with-ttl).
+
+See the [Expiring rows](/docs/concepts/expire-rows/) concept page for
+all modes, worked examples, and semantics (NULLs, ties, monotonicity).
```
javier: worked examples sounds weird to me. It might just me my poor English

documentation/query/sql/create-mat-view.md
```diff
@@ -278,6 +316,62 @@ Time units: `HOURS`, `DAYS`, `WEEKS`, `MONTHS`, `YEARS`
...
+read another policied view, as its base or in a join:
+
+```questdb-sql title="Passthrough view that keeps the latest row per symbol"
+CREATE MATERIALIZED VIEW trades_latest AS (
+  SELECT * FROM trades
+) EXPIRE ROWS KEEP LATEST PARTITION BY symbol;
+```
+
+A `WHEN` predicate is for rules that move with **wall-clock time**, such as a
+rolling `timestamp < dateadd('d', -7, now())` window. The defining query cannot
```
javier: I'd rather use TICK syntax here

documentation/concepts/materialized-views.md
```diff
@@ -12,6 +12,11 @@ their results at query time, materialized views persist their data to disk,
 making them particularly efficient for expensive aggregate queries that are run
```
javier, line 12: General comment for this file, not for this line.

In the performance section of materialized views, we should probably mention the extra queries that are executed when the views  have expiring rows?

javier: Also, we have a whole section on optimizing for latest on, which EXPIRING VIEWS completely made obsolete, right?

documentation/query/sql/create-mat-view.md
```diff
@@ -278,6 +316,62 @@ Time units: `HOURS`, `DAYS`, `WEEKS`, `MONTHS`, `YEARS`
...
+
+A view can carry both `TTL` and `EXPIRE ROWS`. `TTL` comes first in the
+statement and first in effect: it removes rows from the view, and the
+`EXPIRE ROWS` policy then applies to the rows that stay. See
+[Combining with TTL](/docs/concepts/expire-rows/#combining-with-ttl).
+
+See the [Expiring rows](/docs/concepts/expire-rows/) concept page for
+all modes, worked examples, and semantics (NULLs, ties, monotonicity).
+
 ## Complete example
```
javier: The complete example is only for aggregating views. We should probably say complete examples and add an example of each?

Or just remove the complete example

javier: Also, the links at the bottom of this page link to alter view TTL, but not to alter view EXPIRE ROWS

javier (changes requested):

documentation/concepts/expire-rows.md
mtopolnik: `LATEST ON` is pre-existing syntax, and this actually gets rewritten to the existing query. So it makes sense to keep it for familiarity, although I don't know why it's even there -- you MUST name the designated timestamp, it's 100% redundant.

Given that, would you then add `ON` to `KEEP HIGHEST/LOWEST`?

documentation/concepts/expire-rows.md
mtopolnik: Unfortunately, TICK specifies an interval, and here we specify a point in time that is the cutoff point.

Also, syntactically we use a general BOOLEAN predicate, whose one special case is comparing the row's timestamp to `now()`.

To make TICK syntax work here, we'll need new syntax specific to this timeshift predicate.

documentation/query/sql/create-mat-view.md
mtopolnik: It's Opus 5's poor English, don't let it gaslight you :D

documentation/concepts/materialized-views.md
mtopolnik: Added performance notes in 5eac2ec9646ef9750c4797dfb45eea4902b23fb7

As for optimizing LATEST ON, unfortunately row expiry doesn't help here because it still keeps the entire history, and only adds more clauses to the user's query. The feature actually deletes data only for a certain clas of EXPIRE WHEN predicates, those that provably expire all rows before a cutoff point.

documentation/query/sql/create-mat-view.md
mtopolnik: addressed in 98bcfe6acab5fbc8b5e35768d45f22f090dbf678

documentation/concepts/expire-rows.md
javier: We actually had some versions in which ON was optional and you could just do PARTITION BY, but that works only for some queries and I think it is deprecated, so we always want to use the ON. Not sure why. But I would like the consistency in the ROWS KEEPS. It feels wrong that one is ROWS KEEP LATEST ON Column, and the other ROWS KEEP HIGHEST Column

documentation/concepts/expire-rows.md
javier: ah! ok, I thought it was an interval. All good then. We still use dateadd in some queries where TICK doesn't work, so no issue

documentation/concepts/materialized-views.md
javier: Then we can probably add it there, as when I saw we can use EXPIRE ROWS KEEP LATEST ON ts, I thought I could just use that instead of going through the hoops of an aggregating matview. We should explain probably the tradeoff and show the current way of doing it efficiently

mtopolnik: Addressed in bea05b4dd0362b6650b37c6b16797e067b7fc035.

This points out exactly which form of EXPIRE ROWS removes expired rows, and also points out that the pre-aggregation optimization is still relevant, both where it's described, and in a pointer to it from the EXPIRE ROWS documentation.

documentation/concepts/expire-rows.md
javier: I believe this is now the only unresolved comment. Not sure what's the situation of this, if we are going to keep the inconsistency or not

mtopolnik: Addressed in d703c8827aa8c3c29ce7ef2cd268c45932092754

Also made changes to the QuestDB PRs.

javier (approved):

PR #483, opened 2026-07-11
docs: document query.timeout duration key alongside query.timeout.sec

PedroHenrique0713:
## What

Documents the `query.timeout` configuration key on the Cairo engine reference page, next to the existing `query.timeout.sec`, and clarifies how the two relate.

## Why

Part of #274, which reports the query timeout configuration as inconsistent. The root cause is a documentation gap: both keys are valid and related, but no page explains that.

From the source (`PropServerConfiguration.java`):

```java
this.queryTimeout = (long) (getDouble(properties, env, PropertyKey.QUERY_TIMEOUT_SEC, "60") * Micros.SECOND_MILLIS);
this.queryTimeout = getMillis(properties, env, PropertyKey.QUERY_TIMEOUT, this.queryTimeout);
```

- `query.timeout.sec` — a number of seconds (default `60`)
- `query.timeout` — a duration parsed by `Numbers.parseMillis` (e.g. `120s`, `1m`, `500ms`) that overrides `query.timeout.sec` when set

Today the reference page (`configuration/cairo-engine.md`) only lists `query.timeout.sec`, while the configuration overview example uses `query.timeout=120s`. Neither mentions the other, which is what makes the two look inconsistent. This change makes the reference document both keys so the pages agree.

## Scope

This addresses the `query.timeout` point of #274 only. The other two points look out of scope for this docs repo:

- `cairo.auto.scale.symbol.capacity` — its default changed in questdb/questdb#6352, so it is better covered by the broader config refresh mentioned in the issue thread.
- `line.tcp.auth.db.path` — already documented in `configuration/ingestion.md`; the "missing from the default config" note is about the generated `server.conf` template in the core repo, not these docs.

## Testing

Docs-only prose change with no new links or components, so the Docusaurus broken-link build covers it.


javier (changes requested):
I think we should add an entry for the query.timeout property right about query.timeout.sec, and explain in both sections which one has priority.

PedroHenrique0713:
Done, thanks for the review.

`query.timeout` now has its own entry directly above `query.timeout.sec`, and the precedence rule is stated in both sections.

While writing it I noticed `query.timeout.sec` is registered as deprecated in favour of `query.timeout` (`PropServerConfiguration`, `registerDeprecated(QUERY_TIMEOUT_SEC, QUERY_TIMEOUT)`), so the old entry now carries a short deprecation note pointing at the replacement. Both keys feed the same value: `query.timeout.sec` is read first (default 60), then `query.timeout` overrides it when present, which is why the new entry documents `60s` as its effective default. The duration units are the ones `Numbers.parseMillis` accepts (`ms`, `s`, `m`, `h`), with a bare number read as milliseconds.

Happy to drop the deprecation note if you would rather keep that for a separate change.

javier (approved):

javier:
Thanks! publishing it now!

Issue #305, opened 2025-12-24
Web console update needed (export to parquet, views)


javier:
The web console section is not mentioning parquet on the results pane. It will also soon need to add information about views

PedroHenrique0713:
Opened #484 documenting the Parquet download on the Result Grid. The primary download button is now Parquet (`fmt=parquet`), with CSV moved under the dropdown (`fmt=csv`) — verified against the web-console E2E tests (`questdb/ui`, `e2e/tests/console/download.spec.js`).

I left the "views" part out for now, since that UI doesn't seem documented yet. Happy to follow up when it is.

PedroHenrique0713:
Following up on my July note, since I said I would come back when the views UI was documented.

Both halves of this issue are now covered:

**Parquet on the results pane** — `getting-started/web-console/result-grid.md` L24-25 documents the Parquet download as the primary button and CSV under the dropdown. That went in with #484.

**Views** — the web console pages picked this up in the meantime:
- `web-console/schema-explorer.md` covers views and materialized views in the object tree, the base tables shown for a materialized view, the **Create materialized view** action that generates the `CREATE MATERIALIZED VIEW` statement, and the warning icon plus invalidation reason on an invalid view
- `web-console/table-details.md` covers the Table/Materialized View type badge and the metadata panel for either object type
- `web-console/overview.md` L73 and L79 point at both from the top

I did not write the views part, so this is a report rather than a claim. But as far as I can tell nothing from the original request is missing any more, and the issue can be closed.

javier:
Thank you, Pedro!

PR #484, opened 2026-07-12
docs: document Parquet and CSV download options in the Result Grid

PedroHenrique0713:
## What

Updates the Result Grid download action to document both the **Parquet** and **CSV** export options.

## Why

Addresses #305. The Result Grid docs only mentioned CSV, but the current Web Console shows Parquet as the primary download button and moves CSV into a dropdown.

Verified against the Web Console E2E tests (`questdb/ui`, `e2e/tests/console/download.spec.js`):

- `download-parquet-button` is shown directly and requests `fmt=parquet`
- `download-csv-button` appears under `download-dropdown-button` and requests `fmt=csv`

So the previous single "Download result as CSV" bullet was both incomplete and out of date. This lists Parquet as the primary download and notes CSV is available from the dropdown, linking to the [Parquet export](/docs/query/export-parquet/) page.

## Scope

Covers the Parquet part of #305. The "views" note in that issue is left out, since the web-console views UI isn't documented yet ("will also soon need").

## Testing

Docs-only change; the linked `/docs/query/export-parquet/` page exists, so the Docusaurus broken-link build passes.

### DIFF
--- a/documentation/getting-started/web-console/result-grid.md
+++ b/documentation/getting-started/web-console/result-grid.md
@@ -21,7 +21,8 @@ The Result Grid provides several action buttons in the toolbar to help you work 
 - **Move selected column to the front**: Moves the currently selected column to the leftmost position for better visibility
 - **Reset grid layout**: Resets the grid to its default column arrangement and removes all customizations including frozen columns and column reordering
 - **Refresh**: Re-executes the last query to update the results with fresh data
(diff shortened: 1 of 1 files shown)


javier (approved):

javier:
Thanks! Sorry it took me a while to review

PR #508, opened 2026-08-07
Document multi-dimensional numpy arrays in the QWP Python client

tris0laris:
## What

The value types table on the QWP Python client page maps `numpy.ndarray` of `float64` to `DOUBLE[]` and stops there, which reads as though 1D arrays are all that is supported.

N-dimensional arrays work, and land in the matching N-dimensional column type. This adds the missing detail plus a short section with a runnable example.

## Why

Order book snapshots are a headline use case for QWP, and they need 2D. I hit this building an OKX order book pipeline: nothing on the page suggested 2D would work, so I had to test it rather than read it.

## Verified against 10.0.0

| numpy input | auto-created column |
| --- | --- |
| 1D `float64` | `DOUBLE[]` |
| 2D `float64` | `DOUBLE[][]` |
| 3D `float64` | `DOUBLE[][][]` |
| `float32` | rejected, `Only float64 numpy arrays are supported, got dtype: float32` |
| `int64` | rejected, `Only float64 numpy arrays are supported, got dtype: int64` |

The example in the diff was run verbatim against a local 10.0.0 instance. It auto-creates:

```
symbol SYMBOL, bids DOUBLE[][], asks DOUBLE[][], timestamp TIMESTAMP_NS
```

A 2D array also round-trips intact: shape `(2, 25)` in, shape `(2, 25)` numpy array back out of `to_pandas()`.

The dtype restriction is documented too, since it is easy to hit (market data often arrives as `float32` or ints) and was not mentioned anywhere on the page.

### DIFF
--- a/documentation/connect/clients/python.md
+++ b/documentation/connect/clients/python.md
@@ -287,46 +287,6 @@ strings in `columns` become `VARCHAR`. `DECIMAL` columns must be created
 ahead of time with `CREATE TABLE ... (price DECIMAL(18, 2), ...)`; the server
 does not auto-create them.
 
-#### Multi-dimensional arrays
-
-A numpy array keeps its shape on the wire, so an N-dimensional array lands in an
-N-dimensional QuestDB column. This is the natural fit for order book snapshots,
-where one row carries a whole book:
-
-```python
-import numpy as np
-import questdb
-from questdb import TimestampNanos
-
-# shape (2, N): row 0 = prices, row 1 = sizes
-bids = np.array([[64901.6, 64901.5], [3.02, 0.06]], dtype=np.float64)
-asks = np.array([[64901.7, 64901.8], [1.54, 0.21]], dtype=np.float64)
-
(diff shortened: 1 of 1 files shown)


documentation/connect/clients/python.md
```diff
@@ -287,6 +287,46 @@ strings in `columns` become `VARCHAR`. `DECIMAL` columns must be created
 ahead of time with `CREATE TABLE ... (price DECIMAL(18, 2), ...)`; the server
 does not auto-create them.
 
+#### Multi-dimensional arrays
```
javier: I would make this level 3, so it appears on the right index

javier (changes requested):

documentation/connect/clients/python.md
javier:
```suggestion
### Multi-dimensional arrays
```

javier (approved):

PR #208, opened 2025-07-07
Improve Order Book Analytics page

mtopolnik:
- align the schema with `market_data` on Demo
- use `DECLARE` in almost all queries

PR #518, opened 2026-08-18
docs: add Kubernetes Operator manual

sklarsa:
## Summary

- Add an additive-only, 14-page Kubernetes Operator manual under `documentation/operator/`.
- Nest the manual under the existing **Deployment** sidebar category, alongside the other deployment guides.
- Preserve existing deployment documentation and add no redirects.

## Navigation

The Operator sits inside **Deployment → Kubernetes Operator** rather than as a
top-level section, so it does not compete with the primary docs sections for
attention. Within it:

- **Overview** → `operator/index`
- **Installation** — shared prerequisites and the canonical Helm install
- **Getting Started** — Amazon EKS and Azure AKS onboarding guides
- **Configuration**
- **Day 2 operations** — operator, database, backup/restore, high availability, troubleshooting
- **Reference** — known limitations, support, releases, API reference

The cloud guides are reached through a **Getting Started** dropdown rather than a
landing page; the earlier `operator/getting-started/index.md` only linked to the
two guides the dropdown now exposes directly, so it was removed. No other page
linked to it.

## Provenance

- Source operator release: tag `v0.1.0`, commit `db63afd4df91d1eed3af49eca809c99807453d5d`
- Operator generator head: `fcbe034910c0cb34955b497da7eab30993bf7c74`

## Validation

- `yarn build` passes with `onBrokenLinks: "throw"`, confirming the removed
  landing page left no dangling links.
- All 14 operator routes build; the `/operator/getting-started/` landing route is
  gone as intended.
- Operator `docs-sync-check` passed in the prior validation report.

## Companion change

Companion operator PR: https://github.com/questdb/questdb-enterprise-operator/pull/204

## Residual risks

- Non-blocking review note: raw Markdown projections contain duplicate top-level
  headings because pages have both frontmatter titles and explicit H1 headings;
  rendered HTML contains only one title.


documentation/sidebars.js
```diff
@@ -797,6 +796,62 @@ module.exports = {
         "deployment/digital-ocean",
         "deployment/hetzner",
         "deployment/compression-zfs",
+        {
+          type: "category",
```
javier, line 793: I don't like the placement. We have a kubernetes page at the same level, so I think the natural place would be right after. Also, I would probably cross link from both pages to each other

Image

javier: Also, seeing how this is Enterprise only, we might want to rename to Kubernetes Enterprise Operator. Or Enterprise Kubernenetes Operator

documentation/enterprise-kubernetes-operator/getting-started/aws.md
```diff
@@ -0,0 +1,645 @@
+---
```
javier, line 1: Note for all the files. I don't like the idea of using the slug /docs/operator/ for all these pages. It is way too generic. I would rather use /docs/enterprise-kubernetes-operator or similar.

documentation/operator/index.md
```diff
@@ -0,0 +1,38 @@
...
+
+# QuestDB Enterprise Operator
+
+:::warning[Beta for design partners]
+The API is `questdb.io/v1alpha1` and may change between beta releases.
+Amazon EKS and Azure AKS are supported on the tested versions in the
+[support matrix](/docs/operator/known-limitations/#supported-platforms-and-versions).
+See [Support](/docs/operator/support/) for beta expectations.
+:::
+
```
javier: maybe an opportunity to add an info box like we have at tls or rbac, saying this is enterprise only and linking to the enterprise page. See /docs/security/tls/

documentation/operator/getting-started/aws.md
```diff
@@ -0,0 +1,645 @@
...
+
+This guide covers managed Linux node groups. It does not cover Windows,
+Fargate, Hybrid Nodes, EKS Auto Mode, Karpenter, or self-managed node groups.
+
+You'll start with one QuestDB instance. It needs 1 CPU, 2 GiB of memory, and one
+20 GiB EBS volume.
+
+Each additional instance needs the same resources.
+
+For high availability, place worker nodes in more than one Availability Zone.
```
javier: Should we reassure users here that instances do not communicate with each other directly and they only communicate via the S3 shared object store?

documentation/operator/getting-started/azure.md
```diff
@@ -0,0 +1,483 @@
...
+- capacity for one 20 GiB Azure Disk.
+
+This guide covers Linux clusters only.
+
+You'll start with one QuestDB instance. It needs 1 CPU, 2 GiB of memory, and one
+Azure Disk.
+
+Azure may bill the 20 GiB request as a 32 GiB disk SKU.
+
+Each additional instance needs the same resources. For high availability, place
```
javier: same as aws

Should we reassure users here that instances do not communicate with each other directly and they only communicate via the  shared azure blob object store?

documentation/enterprise-kubernetes-operator/operations/operator.md
```diff
@@ -0,0 +1,266 @@
...
+
+With `prometheus.enable=true`, the beta/default ServiceMonitor uses `insecureSkipVerify: true`. Traffic is encrypted and authenticated, but the scraper does not verify the serving certificate.
+
+Chart-managed verified metrics TLS is [not operational in this beta](/docs/operator/known-limitations/#chart-managed-verified-controller-metrics-tls-is-not-operational). Do **not** enable `certmanager.enable=true` as a verified-metrics solution. Customers requiring certificate verification should use a separately reviewed scraper/certificate integration with QuestDB support.
+
+Do not enable `prometheus.enable` until the ServiceMonitor CRD exists. The chart does not create Prometheus, a scraper ServiceAccount, dashboards, or alerts.
+
+## Upgrade the operator
+
+Only the latest beta is supported. Read its release notes before changing the controller or CRDs; `questdb.io/v1alpha1` may have breaking changes.
```
javier: It sounds weird we support a beta. We don't have a stable release yet?

javier: The reference section has a page that mentions beta, but the label is questdb.io/v1alpha1, another pag ethat says the latest stable release is 0.1.0 and another mentioning v1alpha1 only. It feels inconsistent

documentation/enterprise-kubernetes-operator/operations/database.md
```diff
@@ -0,0 +1,299 @@
...
+| --- | --- |
+| `<name>-rw` | Current primary. Use for writes and administration. |
+| `<name>-ro` | Created for store-backed replication. Routes to qualified replicas, but falls back to the primary when no replica qualifies. It is **not** strict replica isolation. |
+| `<name>` | Headless, internal identity/DNS service. It publishes unready addresses and must not be used as a client availability endpoint. |
+
+| Port | Protocol |
+| --- | --- |
+| 9000 | HTTP and Web Console |
+| 8812 | PostgreSQL wire protocol |
+| 9009 | InfluxDB Line Protocol over TCP |
```
javier: First mention I see to ILP over TCP. We should mention this when configuring EKS and AKS if we want to leave it. And in that case, maybe we want to add 9007, which is the default for UDP in QWP, which is disabled by default but can be enabled via https://questdb.com/docs/configuration/qwp/#qwpudpbindto

documentation/enterprise-kubernetes-operator/operations/backup-restore.md
```diff
@@ -0,0 +1,245 @@
+---
+title: Backup and restore
+description: Configure backups and restore Operator-managed QuestDB clusters, including point-in-time recovery.
+---
+
+# Backup and restore
+
+QuestDB Enterprise schedules and performs backups on the primary. The operator configures the engine and observes `backups()`; it does not upload data, create a Kubernetes CronJob, expose an on-demand backup API, or read the object store. Status can lag the database by roughly the manager's resync/observation interval (about two minutes).
```
javier, line 8: should we link to the backups page here?

documentation/enterprise-kubernetes-operator/high-availability.md
```diff
@@ -0,0 +1,269 @@
+---
+title: High availability
+description: Operate replication, planned promotion, emergency failover, and migration with the Kubernetes Operator.
+---
+
+# High availability
+
+Store-backed replication has one primary and zero or more replicas. A new replica is born from a completed backup, then consumes the primary's object-store WAL.
```
javier, line 8: should we link to the replication page?

documentation/enterprise-kubernetes-operator/high-availability.md
```diff
@@ -0,0 +1,269 @@
...
+| Replica is lost | Recreate/re-seed it from backup and WAL when prerequisites are available. |
+
+A `QuestDBPromotion` records one explicit, one-shot cutover and remains as its audit object. If a same-name primary Pod was recreated before its PVC deletion began, do not wait forever for the mounted Terminating claim: current operator versions automatically fence that established Pod so `pvc-protection` can release the PVC. `status.currentPrimary` remains the non-empty primary of record until an explicit promotion; fencing is not automatic failover.
+
+## Promotion and failover
+
+| Mode | Use | Data effect |
+| --- | --- | --- |
+| `Planned` | Healthy primary and live replica | Drains the old primary and verifies the target before promotion. Fails rather than becoming lossy. |
+| `Emergency` | Primary is lost/wedged and service restoration outweighs data loss | Skips the drain. Unreplicated writes are lost. |
```
javier: With the new QWP clients, if we are using the durable ack flag in the connection, the client should automatically retry to send all the data that was in flight in the primary but never reached the WAL, so I believe in this case we would have no data loss. We'd need to check with Vlad or Jaromir https://questdb.com/docs/connect/clients/connect-string/#durable-ack

documentation/enterprise-kubernetes-operator/high-availability.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+Confirm `status.currentPrimary` names the target and require current-generation `Available=True/PrimaryReady`, `Progressing=False/Settled`, and `WriteHealthy=True/Healthy`. Test `<name>-rw` connectivity, critical table timestamps/row counts, and application writes before closing the change. Confirm replicas return healthy after the demoted primary is re-seeded. Finally inspect PVC labels: the target PVC must have `questdb.io/role=primary`, while every demoted/re-seeded replica PVC must have `questdb.io/role=replica`. The separate `questdb.io/bootstrap` label records ancestry/state and does not identify the current role.
+
+## Emergency promotion
+
+Use the same object shape with `mode: Emergency` only after selecting a live target and accepting data loss.
+
+:::danger
+Emergency promotion permanently loses writes that the old primary had not replicated. Clients connected directly to the old pod can briefly receive acknowledgements for writes that will be discarded while it detects loss of store ownership. Stop/repoint writers and use the stable `<name>-rw` Service, not pod addresses.
```
javier: same comment as above about durable acks

documentation/enterprise-kubernetes-operator/reference/api.md
```diff
@@ -0,0 +1,4490 @@
...
+
+- [QuestDBObjectStore](#questdbobjectstore)
+
+- [QuestDBPromotion](#questdbpromotion)
+
+
+
+
+## QuestDBCluster
+<sup><sup>[↩ Parent](#questdbiov1alpha1 )</sup></sup>
```
javier: Not sure what this is doing here

documentation/enterprise-kubernetes-operator/reference/api.md
```diff
@@ -0,0 +1,4490 @@
...
...i>!has(self.bootstrap) || !has(self.bootstrap.follow) || (has(self.replication) && has(self.replication.root) && size(self.replication.root) &gt; 0): spec.bootstrap.follow requires an explicit spec.replication.root naming the external source's WAL prefix. Omitted, the root defaults to this cluster's own identity-scoped db/&#123;namespace&#125;/&#123;name&#125;/, which is a stream nothing is writing to — the follower would restore and then never advance</li><li>!has(self.bootstrap) || !has(self.bootstrap.follow) || (has(self.backup) && has(self.backup.root) && size(self.backup.root) &gt; 0): spec.bootstrap.follow requires an explicit spec.backup.root naming the external source's backup prefix. Omitted, it defaults to a bare 'backup/' that is not scoped to any cluster, so the follower would look for its baseline wherever that happens to point</li><li>!has(self.bootstrap) || !has(self.bootstrap.follow) || !has(self.replication) || !has(self.replication.seedFrom): spec.replication.seedFrom is meaningless under spec.bootstrap.follow: a follower's replicas seed from the external source's backup named by follow.sourceInstanceName, not from an instance of this cluster. Remove seedFrom</li>
```
javier: very weird formatting of the LI bullet points on all the tables in this page

Image

javier (changes requested):
Very cool, but I have some suggestions

documentation/enterprise-kubernetes-operator/operations/operator.md
sklarsa: A lot of k8s resources have "v1alpha1" apis. Maybe I should just drop the "in beta" stuff altogether? It's only beta because no one is really using it...

documentation/enterprise-kubernetes-operator/reference/api.md
sklarsa: I'll remove it for each top-level CRD

documentation/sidebars.js
sklarsa: Moved: the category now sits directly after `deployment/kubernetes` in the Deployment group, instead of at the bottom. Order is Docker → Kubernetes → Enterprise Kubernetes Operator → systemd → AWS → …

Cross-linked both ways:
- `deployment/kubernetes.md` gets a tip after the intro saying that page covers the self-managed Helm chart, and pointing to the operator for Enterprise.
- The operator overview points back at the Helm chart page for open source QuestDB (inside the new Enterprise box, see the other thread).

Also renamed to **Enterprise Kubernetes Operator** per your follow-up. The index H1 said "QuestDB Enterprise Operator" — missing "Kubernetes", and out of sync with its own frontmatter title — so that now matches too.

I left the sub-page titles alone (`Install the Kubernetes Operator`, `Kubernetes Operator known limitations`, …). They read fine under the renamed parent, but they are what shows up in Algolia results, so happy to prefix them all with "Enterprise" if you would rather.

documentation/enterprise-kubernetes-operator/getting-started/aws.md
sklarsa: Agreed, `/docs/operator/` is too generic. Renamed the directory to `documentation/enterprise-kubernetes-operator/`, so the pages are now under `/docs/enterprise-kubernetes-operator/…`, matching the new sidebar label.

Rewrote all 30 internal `/docs/operator/` links and the 14 sidebar doc IDs. Nothing outside the operator directory and `sidebars.js` referenced the old path, and no `_redirects` entry is needed since these pages have never shipped.

`yarn build` passes with `onBrokenLinks: "throw"`, so every rewritten link resolves.

documentation/operator/index.md
sklarsa: Added the same `<EnterpriseNote>` component that `/docs/security/tls/` and `/docs/security/rbac/` use, above the beta warning. It carries the "Learn more →" link to questdb.com/enterprise built into the component, and I used the body text to point open source users at the Helm chart page — which doubles as the reverse cross-link you asked for in the other thread.

documentation/enterprise-kubernetes-operator/reference/api.md
sklarsa: Fixed in the generator: questdb/questdb-enterprise-operator#206.

crdoc emits validation items as bare `<li>` with no list around them, so browsers rendered the markers with no list box and no indent — that is the overflow in your screenshot. The sanitizer now wraps each run in `<ul>`, and `api.md` is regenerated: all 20 runs wrapped, rendering as a real nested list inside the cell.

documentation/enterprise-kubernetes-operator/reference/api.md
sklarsa: Removed in the generator: questdb/questdb-enterprise-operator#206.

crdoc gives each top-level resource a `[↩ Parent]` link back to its package heading, through a padded anchor (`#questdbiov1alpha1 ` — note the trailing space) that navigates nowhere. The sanitizer now drops those. The 78 nested backlinks are kept, since those do go somewhere.

Regenerating `api.md` produced exactly 20 insertions and 23 deletions: the 20 bullet lines plus these 3 backlinks, nothing else.

documentation/enterprise-kubernetes-operator/high-availability.md
sklarsa: Added — it now mirrors the pointer the backup and restore runbook already carries to `/docs/operations/backup`:

> For more information about replication mechanics, please refer to the [official documentation](/docs/high-availability/overview/).

Same split as backup: the Operator page covers the Kubernetes mechanics (promotion, failover, migration) and defers how replication actually works to the engine documentation.

documentation/enterprise-kubernetes-operator/operations/database.md
sklarsa: Turned this into CRD flags rather than documentation: questdb/questdb-enterprise-operator#208.

Optional protocols are now spec-driven, so enabling a listener and publishing its port are one decision:

```yaml
spec:
  protocols:
    ilpTCP:
      enabled: false    # default — 9009 closed unless asked for
    qwp:
      udp:
        enabled: false  # default — 9007 closed unless asked for
```

ILP/TCP defaults to **off**, deliberately the opposite of the engine default, so an operator-managed cluster does not open an unauthenticated ingestion socket by accident. That answers the "if we want to leave it" half — it is now opt-in, and the port is only published on the Services while the listener is enabled.

On QWP: there is no on/off switch for QWP over WebSocket. `/write/v4` and `/read/v1` are served by the HTTP server on 9000 and share its network settings, so they are on whenever HTTP is. The only QWP toggle the engine has is `qwp.udp.enabled`, which is the 9007 you pointed at — so that is the only one exposed.

Leaving this thread open: the docs half is still to do. `configuration.md` needs a section for the new block, and the EKS/AKS guides should cover the ports once #208 ships in a release the API reference regenerates from.

documentation/enterprise-kubernetes-operator/operations/operator.md
javier: I would just say only the latest version and get rid of beta, yes

documentation/enterprise-kubernetes-operator/operations/database.md
javier: Looking good. Leaving open until we have the new config block

documentation/enterprise-kubernetes-operator/index.md
```diff
@@ -0,0 +1,46 @@
...
+
+# QuestDB Enterprise Kubernetes Operator
+
+<EnterpriseNote>
+  The Kubernetes Operator manages QuestDB Enterprise clusters. For open source
+  QuestDB on Kubernetes, use the
+  [QuestDB Helm chart](/docs/deployment/kubernetes/).
+</EnterpriseNote>
+
+:::warning[Beta for design partners]
```
javier: Hadn't noticed the beta reference here before. Should we just say Alpha all over?

documentation/enterprise-kubernetes-operator/operations/database.md
sklarsa: Done — the docs half has landed, and the operator flags shipped in [v0.2.0](https://github.com/questdb/questdb-enterprise-operator/releases/tag/v0.2.0) (questdb/questdb-enterprise-operator#208).

**Ports table** now says what is always on and what is opt-in:

| Port | Protocol | Published |
| --- | --- | --- |
| 9000 | HTTP, Web Console, QWP over WebSocket | Always |
| 8812 | PostgreSQL wire protocol | Always |
| 9009 | InfluxDB Line Protocol over TCP | Always |
| 9003 | Min HTTP server: health and metrics | Always |
| 9007/UDP | QWP UDP receiver | Only when enabled |

**EKS and AKS guides** now separate the paths the operator itself needs from the ports ingestion and query clients use, and link to the full table — that was the gap you spotted, ILP/TCP appearing first in the day-2 runbook when the reader decides what to allow during setup.

**On QWP:** there is no on/off switch for QWP over WebSocket. `/write/v4` and `/read/v1` are served by the HTTP server on 9000 and share its network settings, so they are on wherever 9000 is. `qwp.udp.enabled` — your 9007 — is the only QWP toggle the engine has, so that is the one exposed, as `spec.protocols.qwp.udp.enabled`, defaulting off.

Configuration gains a [Wire protocols](https://questdb.com/docs/enterprise-kubernetes-operator/configuration/#wire-protocols) section. Two things it calls out before you enable the receiver: it is **unauthenticated** (QWP authenticates on the WebSocket upgrade request and UDP has no upgrade, so anything that can reach the port can write — restrict with a NetworkPolicy), and delivery is fire-and-forget. It is also not published on `-ro`, since a datagram aimed at a replica is discarded with nothing returned to say so.

I did not add a flag for ILP over TCP. Defaulting it off would close 9009 on existing clusters and reject `line.tcp.enabled` in `spec.config`, so it is worth its own PR and its own decision rather than riding along with this one.

documentation/enterprise-kubernetes-operator/index.md
sklarsa: Why won't we just remove the beta/alpha from the docs altogether? Keep the api version because that's a common k8s convention, but just release it "as is"?

javier: sounds good to me

sklarsa: Done in 7420952 — the beta/alpha framing is gone from the manual, `questdb.io/v1alpha1` stays.

The admonition on this page is now `:::warning[Supported platforms]` and says the API "may change between releases". Elsewhere: `support.md` gains a **Support scope** section in place of **Beta support**, `known-limitations.md` an **API lifecycle** section in place of **Beta lifecycle**, and installation/operator/backup-restore lose the remaining "in this beta" qualifiers. Nothing linked to the two renamed anchors and these pages have never shipped, so no redirects.

The two generated pages needed the source fixed rather than the output: questdb/questdb-enterprise-operator#210 drops "beta-supported wiring" from the `credentialsSecret` CRD description and rewrites the 0.1.0 CHANGELOG note. I re-ran `sync-docs.sh`'s pipeline against that branch and took the diff, so `reference/api.md` and `releases.md` here match exactly what the next release will regenerate — that closes the "one page says latest stable, another says beta" inconsistency you spotted in the other thread.

One thing I left: the API reference still carries "This is a beta field and requires the MatchLabelKeysInPodTopologySpread feature gate". That is upstream Kubernetes text about a real k8s feature gate on `topologySpreadConstraints`, not us — crdoc re-emits it every regeneration.

documentation/enterprise-kubernetes-operator/operations/operator.md
sklarsa: Done in 7420952 — agreed on both counts, and the inconsistency you spotted across the reference section is closed.

This page now reads "Only the latest release is supported. Read its release notes before changing the controller or CRDs; `questdb.io/v1alpha1` may have breaking changes." The upgrade table row is "The release changed a CRD schema…" rather than "a beta schema", and the metrics section drops "the beta's default lack of certificate verification" and "not operational in this beta".

On the three-way inconsistency — beta prose, a `v1alpha1` label, and a page calling 0.1.0 the latest stable — the fix had to be upstream, since `reference/api.md` and `releases.md` are generated from the Operator repo, not written here. questdb/questdb-enterprise-operator#210 removes "beta-supported wiring" from the `credentialsSecret` CRD description and rewrites the 0.1.0 CHANGELOG note to "Initial release". I re-ran `sync-docs.sh`'s pipeline against that branch and took the diff, so both generated pages here already match what the next release regenerates. The whole manual now says one thing: latest release, `v1alpha1` API, no beta.

documentation/enterprise-kubernetes-operator/high-availability.md
sklarsa: Updated in c928075. Emergency promotion now says unreplicated writes may be lost and notes that QWP clients using durable acknowledgements retain and retry writes that were not durably acknowledged. The table links directly to the durable-ack guidance.

documentation/enterprise-kubernetes-operator/high-availability.md
sklarsa: Updated in c928075 with the same durable-ack qualification and link: clients retain writes until object-store durability is confirmed and retry unacknowledged writes after connecting to the new primary.

javier (approved):

PR #523, opened 2026-08-24
docs(dotnet): document 4.0.0 release and zstd assembly split

ideoma:
## Summary
- Documents questdb/net-questdb-client#81: the optional `net-questdb-client-zstd` package for `QueryClient` egress compression, including the fail-fast `ConfigError` for explicit `compression=zstd` when it's absent, `compression=auto`'s silent degrade to raw, and the trimmed/Native AOT reflection caveat.
- Bumps the beta caution's version reference from `4.0.0-beta.1` to `4.0.0` per questdb/net-questdb-client#82.

### DIFF
--- a/documentation/connect/clients/dotnet.md
+++ b/documentation/connect/clients/dotnet.md
@@ -36,7 +36,7 @@ services. Construct one, share it across threads, dispose it at shutdown.
 :::caution Beta
 
 The QWP sender, `QueryClient`, and store-and-forward APIs on this page are
-**beta**, shipped in `4.0.0-beta.1`. Expect the API to change before GA; track
+**beta**, shipped in `4.0.0`. Expect the API to change before GA; track
 the [client releases](https://github.com/questdb/net-questdb-client/releases).
 The ILP over HTTP and TCP transports are unaffected and remain stable.
 
@@ -68,6 +68,11 @@ Install the NuGet package with the dotnet CLI:
 ```shell
 dotnet add package net-questdb-client
 ```
+
+`net-questdb-client` has no zstd dependency. To use `compression=zstd` on the
+`QueryClient` egress path, also add the optional
+[`net-questdb-client-zstd`](https://www.nuget.org/packages/net-questdb-client-zstd/)
+package. See [Compression](#compression).
 
 ## Quick start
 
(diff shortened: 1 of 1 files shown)


ideoma:
## Review (level 0)

Reviewed at level 0 — single-page, 40-line diff to an existing page, no sidebar or link-topology changes.

**Scope:** `documentation/connect/clients/dotnet.md` only — 34 insertions/6 deletions. No new page, no `sidebars.js` change, no other page references `net-questdb-client`.

**Verification performed:** cross-checked every technical claim against the upstream `questdb/net-questdb-client` repo (issue #81/#82, merged), the actual `net-questdb-client-zstd.csproj`, `QwpZstdCodec.cs` source, and NuGet:
- `net-questdb-client-zstd` package exists on NuGet at `4.0.0` ✓
- Main `net-questdb-client.csproj` carries no `ZstdSharp.Port` reference (confirms "no zstd dependency") ✓
- Plugin `TargetFrameworks` is `net7.0;net8.0;net9.0;net10.0`, matching the "targets net7.0 and up" claim ✓
- `QwpZstdCodec.Create()` throws exactly `new IngressError(ErrorCode.ConfigError, ...)` when the plugin can't be resolved ✓
- `compression=auto` path (`IsAvailable` false → factory null, no throw) matches the "silently connects as raw" claim ✓
- Plugin loaded via `Assembly.Load` + `[DynamicDependency]`/`[RequiresUnreferencedCode]`/`[RequiresDynamicCode]` attributes — supports the trimming/AOT caution ✓
- GitHub release `4.0.0` is tagged "Latest" (2026-08-20), confirming the beta→4.0.0 version bump ✓
- `yarn build` passes clean, no broken links ✓

### Verdict: Approve

### Non-blocking
- **Minor/nit:** "RESULT_BATCH" (line 1426) is used without being defined anywhere on the page or elsewhere in the docs — it's inferable from context, but a reader/LLM with no QWP protocol background has to guess it means "query result batch." Not worth blocking on.
- No em-dashes, no LLM-mannerism filler, admonition style (`:::note Trimmed and Native AOT publishes`) matches the page's existing `:::caution`/`:::tip` convention.

### What works
- Every specific claim (exception type, package name/version, target frameworks, fallback behavior) traces directly to the linked upstream source rather than being asserted.
- The new compression table's "with plugin / without plugin" split cleanly answers the actual failure mode instead of just describing the happy path.
- Forward/back links between the install note and the `#compression` section keep the two additions self-contained without duplicating explanation.

ideoma:
## Review (level 2 — `-L3` requested; skill's max defined level is 2)

Scope: single existing page, no sidebar change (`documentation/connect/clients/dotnet.md`, 52 insertions / 6 deletions).

### Verdict: Approve

### Blocking issues
None. Every technical claim was verified against the real `questdb/net-questdb-client` source and NuGet, not just read for plausibility:
- `EnsureZstdAvailableIfRequired()` really does run before any address is dialled (confirmed in `QwpQueryWebSocketClient.cs`) — matches "fails fast, before any address is even dialled."
- `4.0.0-beta.1`'s csproj really did bundle `ZstdSharp.Port` directly; `4.0.0`'s csproj dropped it in favor of the optional `net-questdb-client-zstd` plugin — confirms the stated breaking change.
- Both packages exist on NuGet at version `4.0.0`.
- The `[DynamicDependency]` trimming mechanism and `net7.0+` plugin target match the source exactly; the base package also declares `<IsAotCompatible>true</IsAotCompatible>`, backing the Native AOT claim.
- `yarn build` passes clean.

### Non-blocking
- **Minor** — "Client installation" (line 72-75) links to `net-questdb-client-zstd` on NuGet but doesn't show the `dotnet add package net-questdb-client-zstd` command, breaking the page's own convention of pairing every package mention with a fenced CLI command (see line 68-70).
- **Minor** — Wording drifts across three spots describing when the `ConfigError` fires: "thrown once connecting starts" (line 866) vs "before connecting" (line 1440) vs "before any address is even dialled" (line 1447). Technically consistent once you read all three, but the middle phrasing is looser than the other two. Suggest aligning line 1440 to "before any address is dialled."
- **Minor** — The "Trimmed and Native AOT publishes" note (1459-1477) sits after the unrelated `compression_level`/`NegotiatedCompression` paragraph instead of right after the "absent or unusable" plugin-availability prose it belongs with.
- **Pre-existing, not introduced by this PR** — `compression_level`'s default (`1`) is never stated on this page (only range + clamp); the page explicitly defers full defaults to `connect-string.md`. Also pre-existing: "### Synchronous errors" is nested under "## Asynchronous error handling," a naming mismatch the new zstd-`ConfigError` row inherits.
- **Nit** — Could note that `auto`'s silent fallback to raw is worth asserting on via `NegotiatedCompression` in a post-publish smoke test.

### What works
Accurate, source-verified technical writing; correct indirection (brief pointer in "Client installation" → full detail in "Compression," no duplication); the 3-column plugin-availability table is a clean way to show the `raw`/`zstd`/`auto` behavior split without hand-waving.

### Downgraded (false positives)
- Adversarial pass called the Native AOT/trimming claim "unverifiable" — dismissed; `IsAotCompatible=true` in the library's own `.csproj` plus the `[DynamicDependency]` attribute directly support it.
- Adversarial pass called the "connecting" timing wording a "contradiction" — downgraded to the Minor wording-precision note above; the underlying sequence is verified consistent.

---
2 agents run, 1 manual deep-verify pass; 5 findings kept (3 Minor + 2 pre-existing notes), 1 Nit, 2 downgraded.

javier (approved):

